PR #5, opened 2015-01-06
Add CSS for selection buttons

tombye:
Adds the version of selection buttons used on Digital Marketplace projects.

See https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/4 for the documentation.


quis:
Are there other versions of the selections buttons used elsewhere? Is there a good reason to have our own?

tombye:
Yes, in govuk_elements. We were using a version with a few differences but there's no reason we can't take the version from elements and add our changes further down the cascade. I'll close this and the other until done.

PR #6, opened 2015-01-07
Extend placeholder selectors, not classes

quis:
The behaviour of `@extend` in SASS can unexpectedly:
- result in properties getting applied more generally than the syntax would suggest
- result in bloated CSS

This can be mitigated by only ever extending placeholder selectors (eg `%foo` rather than `.foo`).

[More explantion in this blog post](http://csswizardry.com/2014/01/extending-silent-classes-in-sass/).

Also worth noting that SCSSLint has an [option that warns about extending classes](https://github.com/causes/scss-lint/blob/master/lib/scss_lint/linter/README.md#placeholderinextend).

### DIFF
--- a/scss/_notification-banners.scss
+++ b/scss/_notification-banners.scss
@@ -2,7 +2,8 @@
 @import "_digital-marketplace-colours.scss";
 @import "_typography.scss";
 
-.banner {
+.banner,
+%banner {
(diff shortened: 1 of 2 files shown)


tombye:
Reads better to in my opinion.

PR #11, opened 2015-01-08
Add support for older versions of Internet Explorer

tombye:
This adds the support for older version of IE built into www.gov.uk projects to the pages with an explanation of how to implement it.


quis:
Made a few wording changes in cea7526, if you're happy with those changes then LGTM :+1:

PR #12, opened 2015-01-09
Add multi-selects

tombye:
Adds the multi-selects used on https://www.gov.uk/cma-cases and Digital Marketplace for the filtering of search results.

An example page has been added on this pull request: https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/13


quis:
Should we be thinking about package management for front end vendor scripts (eg jQuery, Hogan)?

tombye:
They should at least be renamed to reflect their versions. I'd prefer to keep package management out of the actual code (on the `master` branch) and as long as the versions are clear, that should make implementation a matter of specifying a version in whatever dependancy management system is on a project.

quis:
I don't think we should copy across tests for code that we haven't written.

PR #13, opened 2015-01-09
Add multi selects page

tombye:
Add example pages for multi-selects:

mulit-selects screenshot

Includes some changes to the grid pages.

See https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/12 for details of the code changes.


quis:
Seems to be missing [this image](https://github.com/alphagov/finder-frontend/blob/master/app/assets/images/toggle-sprite-retina.png) from the assets.

I have a more general question about image assets as well, but will save that for somewhere else…

PR #14, opened 2015-01-12
Extend question headings

tombye:
Extends the existing `question-headings.scss` file to include code for the question and answer formatting developed as part of the Supplier Submission Portal. (now called `questions.scss`.)

See https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/15 for an example.

Also includes some code changes missed from previous pull requests.

### DIFF
--- a/scss/forms/_questions.scss
+++ b/scss/forms/_questions.scss
@@ -1,0 +1,29 @@
+@import "_colours";
+@import "_typography.scss";
+
+.question {
+  margin: 15px 0 0 0;
+  clear: both;
+}
+
+.first-question {
+  margin-top: 0;
+}
+
+.question-heading {
+  display: block;
+  @include heading-24;
+  font-weight: bold;
+  margin: 0 0 10px 0;
+}
+
+.question-heading-with-hint {
+  @extend .question-heading;
+  margin-bottom: 0;
+}
+
+.question-description {
+  @include copy-19;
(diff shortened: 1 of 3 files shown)


scss/forms/_questions.scss
```diff
@@ -0,0 +1,29 @@
...
+@import "_typography.scss";
+
+.question {
+  margin: 15px 0 0 0;
+  clear: both;
+}
+
+.first-question {
+  margin-top: 0;
+}
```
quis: Should we use `@extend` here? I.e. in the markup, which is preferable:
1. `class="question first-question"`
2. `class="first-question"`

If 2., then I think for consistency `question-first` is more in keeping with `banner-warning`, `banner-success` etc.

scss/forms/_summary.scss
```diff
@@ -19,6 +19,10 @@
   @include box-sizing(border-box);
 }
 
+.first-summary-item-heading {
```
quis: Same query as here: https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/14/files#r22858140

scss/forms/_questions.scss
tombye: I've amended the naming for consistency. I think the use of `@extend` should be implemented here but the same should be done to other parts of the CSS so I'm going to do that in a separate pull request.

PR #18, opened 2015-01-15
Add CSS styleguide

tombye:
Add a guide for naming in selectors (ids and classes) to the README in ./scss.

This involved some changes to the CSS we've built already to make a consistent reading experience across the codebase. (Note: Code brought in from other projects isn't required to conform to the styleguide.)


scss/README.md
```diff
@@ -38,3 +38,77 @@ $mobile-ie6: false;
 
 @import "index.scss";
 ```
+
+## Styleguide
+
+### Naming
+
+The naming of classes and ids normally follows one of these patterns:
+
```
quis: Should we make it explicit that the simplest pattern is just **thing**, e.g. `button`, `question`?

PR #20, opened 2015-01-30
Remove the github.io asset path from breadcrumbs

tombye:
Because `_breadcrumb.scss` is to be included directly into projects, this meant they were using github to serve any images referenced in the CSS.

This splits out where that path is set so it can be relative to the project rather than this code.

This also includes some fixes for the use of paths in multi-select SCSS.


quis:
We should think about including the images in this repository. They are as much an aspect of a component as the CSS or Javascript are. It should then be the responsibility of each project to copy them somewhere as part of a build process.

PR #22, opened 2015-02-04
Fix display of headings in summary tables

quis:
These headings are only meant to be for screen readers. They are not meant to be shown visually, but at the moment they are:

screen shot 2015-02-04 at 14 58 55

This commit hides them.

*P.S. When I make the pull request on the Marketplace I'll improve the language from key/value!*


tombye:
LGTM. Can you put this change into a separate pull request on the gh-pages branch too?

quis:
Done as #23

PR #24, opened 2015-02-06
Columns in summary tables fixed to 50% width

quis:
Need to stop this from happening:
image

PR #25, opened 2015-02-09
Use class naming conventions on summary page

quis:
The naming of classes used in the markup did not quite fit with the [naming convention](https://github.com/alphagov/digitalmarketplace-frontend-toolkit/tree/master/scss#naming).

This meant that the summary page in the toolkit was not getting the CSS applied:

image

Now it is:

image

Also brings in the change to the summary page from #24.

### DIFF
--- a/forms/summary.html
+++ b/forms/summary.html
@@ -138,7 +138,7 @@
     <h2 class="summary-item-heading">
         Summary item
     </h2>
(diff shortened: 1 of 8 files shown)


quis:
This PR is out of date if #26 gets merged. I would favour closing it and making a new request against master.

PR #26, opened 2015-02-10
Generate pages from master

quis:
I propose that it would be better for future pull requests to have all proposals made against one branch (`master`). This pull request changes the process of building the documentation to enable that.

On a technical level it does three main things:
- Generate the documentation entirely using scripts, assets, and markup kept on the master branch
- Rewrites the scripts in `./build_tools` to output the generated pages into the `./pages` folder
- Adds a script to clean the `gh-pages` branch, replace it with the (generated) contents of the `./pages` folder on master and force push that to origin.

WARNING: Running the `push_changes_to_github` script will overwrite the contents of the `gh-pages` branch on origin. I have pushed [a `gh-pages-backup` branch](https://github.com/alphagov/digitalmarketplace-frontend-toolkit/tree/gh-pages-backup) to alphagov/digitalmarketplace-frontend-toolkit as mitigation.


pages_builder/push_to_github_pages.sh
```diff
@@ -0,0 +1,35 @@
+current_branch=$(git rev-parse --abbrev-ref HEAD)
```
tombye: Worth setting the 'e' option with `set -e` to ensure the script exits if it hits any errors.

tombye: Also worth putting `#!/bin/sh` at the start of the file.

quis: Addressed in e40c5aad and at https://github.com/alphagov/digitalmarketplace-frontend-toolkit/commit/9bbe884f1a0554232b9705fb774281bbda5b8662#diff-a1fc7d98932ecd188caf8f5fd6414159R1

quis:
@tombye 9bbe884f hopefully addresses the discussion we just had about preserving history in the `gh-pages` branch.

PR #29, opened 2015-02-19
Add a VERSION.txt file to track the tagged version

tombye:
This will allow tracking of the tagged version from within the release code.


tombye:
I've checked with @quis and he said :+1: so merging.

PR #30, opened 2015-02-19
Add the version to the main index page when built

tombye:
This was previously part of the task that pushed to Github pages, now part of the task that builds the pages.

Includes a fix to the README docs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ pip install -r ./pages_builder/requirements.txt
 ### Generate the pages locally
 
 ```
-python build_tools/generate_pages.py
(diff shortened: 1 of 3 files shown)


quis:
The only concern that I have: If you're working locally then the generated pages will display 'version 1.0.2'. What you will be seeing in the browser is `v1.0.2` **plus** the effects of your local changes. Undecided as to whether that is an issue.

tombye:
I think we have to bear it in mind for when we next add a feature and make a call on it then.

quis:
LGTM :+1: in that case. Preparing to merge…

PR #33, opened 2015-02-20
Add 'search result' as a new component

quis:
This is mostly copying over the existing pattern from the Digital Marketplace frontend app.

The main changes are:
- Bring the class naming conventions in line
- Tighten up the spacing and remove the keylines between elements, as per the prototype

From the Marketplace | In the toolkit 
---|---
screen shot 2015-02-20 at 10 35 43 | screen shot 2015-02-20 at 10 59 15

Prototype:
screen shot 2015-02-20 at 11 04 31


pages_builder/pages/search-result.yml
```diff
@@ -0,0 +1,96 @@
...
+              by Knight Technology Ltd.
+            </p>
+            <p class="search-result-excerpt">
+              Example markup can be found in the
+              <span class="search-result-highlighted-text">pages folder</span>.
+              It is used to generate the example pages for documentation into the
+              pages folder. Generating the documentation pages. The
+              documentation pages can be generated by following these steps.
+            </p>
+            <ul class="search-result-tags">
```
tombye: Seems a little odd that the fact this list isn't identified as tags. Can we bake this information into it?
This could be done by adding a visually hidden heading or by using the `aria-label` to give them a label of 'Tags'.

pages_builder/pages/search-result.yml
```diff
@@ -0,0 +1,96 @@
...
+  </div>
+  <main role="main" id="content" class="wrapper">
+    <div id="wrapper">
+      <header class="page-heading">
+        <h1>Search result</h1>
+      </header>
+      <div class="grid-row">
+        <div class="two-third-column">
+          <div class="search-result">
+            <h1 class="search-result-title">
```
tombye: These search results headings should be h2s I think as we already have a h1 for this page.

quis: Addressed by https://github.com/alphagov/digitalmarketplace-frontend-toolkit/commit/98c6b44af6d5d745ace8b4ad66e1279d5223edab

pages_builder/pages/search-result.yml
quis: @tombye See: https://github.com/alphagov/digitalmarketplace-frontend-toolkit/commit/e28220f9df34b6b2bd24f8379fe201bc9d438043. Tested with Voiceover and it reads sensibly.

Also tried a visually hidden span, e.g. `<span>Framework: </span> G-Cloud 6`. The effect is much the same, but the markup and CSS isn't as clean.

tombye:
https://github.com/alphagov/digitalmarketplace-frontend-toolkit/commit/fc2ff3f0e35c7a1deea31b341ed38f14858d9acf looks like a sensible solution until we have a better understanding of how to describe the meta data to users.

PR #37, opened 2015-02-23
Correct use of path in search result

tombye:
The `$path` variable was being set to different values in the CSS for the multi-select and search-result pages.

This corrects that usage so they are both the same (meaning both can be used on a project that sets `$path` once.


quis:
LGTM :+1:

PR #39, opened 2015-03-02
Add validation as a new pattern

quis:
Based on the code used in the SSP.

image

### DIFF
--- a/pages_builder/pages/forms/validation.yml
+++ b/pages_builder/pages/forms/validation.yml
@@ -35,10 +35,10 @@ content: >
           <label for="example-textbox" class="question-heading">
             What was the name of your first pet?
           </label>
-          <p class="validation-message">
+          <p class="validation-message" id="error-example-textbox">
             This question requires an answer
           </p>
-          <input class="text-box" id="example-textbox" type="text" />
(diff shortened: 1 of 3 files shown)


pages_builder/pages/forms/validation.yml
```diff
@@ -0,0 +1,56 @@
...
+      </div>
+      <div class="validation-wrapper">
+        <div class="question">
+          <label for="example-textbox" class="question-heading">
+            What was the name of your first pet?
+          </label>
+          <p class="validation-message">
+            This question requires an answer
+          </p>
+          <input class="text-box" id="example-textbox" type="text" />
```
tombye: Following advice from Léonie, we should add `aria-describedby` to the `<input>` linking it to the `validation-message` paragraph.

tombye:
Think this needs linking to from the index page.

quis:
Link from forms/index added in e5ead43.

pages_builder/pages/forms/validation.yml
quis: Added in 935d7ba.

tombye:
One thing I forgot to mention: can we wrap it in a form so it's clear the markup that should be used?

quis:
@tombye  d447cb2, good idea

PR #41, opened 2015-03-09
Add 'document' as a component

quis:
## A component for linking to a downloadable file, or a list of such links

This is mostly a lift-and-shift from the Grails app, with some slightly more sensible naming of classes.

image

### Graceful degredation

IE6 | IE 8 | IE 9
--- | --- | ---
image | image | image


toolkit/scss/_document.scss
```diff
@@ -0,0 +1,73 @@
...
+}
+
+%document-link,
+.document-link {
+  @include core-19;
+  margin: 5px 0;
+  padding: 10px 0 10px 0;
+  position: relative;
+  display: block;
+  width: 100%; // Trigger HasLayout on older IEs
```
tombye: I think we need these comments in the final CSS so maybe use `/* */` style?

toolkit/scss/_document.scss
```diff
@@ -0,0 +1,73 @@
...
+  display: block;
+  position: absolute;
+  top: 51%;
+  left: 0;
+  background: transparent;
+  text-transform: lowercase;
+  text-decoration: none;
+
+  &:before {
+    content: '.'; // Insert this with CSS so that it doesn't get read out
```
tombye: Same as the above for the style of comments.

toolkit/scss/_document.scss
quis: Comments amended in 8b78170

quis:
@tombye Made some changes for older versions of IE in 2cace49. See revised screenshots above.

PR #43, opened 2015-03-11
Add option to serve pages from a subfolder

quis:
**This fixes a bug where we have no app CSS on Github Pages.**

We now have absolute paths for all assets served from `pages/public`.

By default, running `python -m SimpleHTTPServer 8000` in the `pages/` directory will JustWork™.

If the root directory is different (eg on Github pages) you can set an optional environment variable (`ROOT_DIRECTORY='/digitalmarketplace-frontend-toolkit'`) which gets prepended to the path to any CSS file. The [Jenkins job](http://37.26.90.26:8080/job/Publish%20toolkit%20documentation/configure) has already been configured to provide this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,8 @@ The documentation can be accessed at [http://alphagov.github.io/digitalmarketpla
 
 ### Static assets
 
-The pages use `pages` as the root directory so you can include any assets from `./pages/public/javascripts` or `./pages/public/images` directly.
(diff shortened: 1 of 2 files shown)


quis:
This branch is currently pushed to Github Pages from local (d8383ca99e532771ff8e61001e906b8d80977630) so you can see it all working: http://alphagov.github.io/digitalmarketplace-frontend-toolkit/

PR #47, opened 2015-03-16
Make search result look closer to GOV.UK finders

quis:
In reviewing the design of the Digital Marketplace it was pointed out that search results could make use of the existing 'finder' pattern on GOV.UK (eg
https://www.gov.uk/drug-device-alerts) which would increase the information density of the page.

We will also be using this pattern on forthcoming pages, eg supplier profile, so now is a good time to change it.

This commit adjusts the `.search-result` to look like a GOV.UK finder, if a GOV.UK finder had a line for 'supplier'.

Before | After
--- | ---
image |  image

### DIFF
--- a/toolkit/scss/_search-result.scss
+++ b/toolkit/scss/_search-result.scss
@@ -4,24 +4,34 @@
 @import "_shims.scss";
 
 .search-result {
-  margin: 0;
-  padding: 0 0 $gutter 0;
+
+  padding: 0 0 $gutter/2;
+  margin: 0 0 $gutter/2;
+
+  border-bottom: 1px solid $border-colour;
+
 }
 
 .search-result-title {
-  @include heading-24;
+
+  @include core-19;
+  font-weight: bold;
   margin: 0;
+
+  a {
+    text-decoration: none;
+  }
+
 }
 
 .search-result-supplier {
-  @include copy-19;
-  color: $secondary-text-colour;
-  margin: 0;
+  @include core-16;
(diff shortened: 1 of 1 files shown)


toolkit/scss/_search-result.scss
```diff
@@ -4,24 +4,34 @@
...
+  margin: 0 0 $gutter/2;
+
+  border-bottom: 1px solid $border-colour;
+
 }
 
 .search-result-title {
-  @include heading-24;
+
+  @include core-19;
```
tombye: Could just use the `bold-19` mixin here: https://github.com/alphagov/govuk_frontend_toolkit/blob/master/stylesheets/_typography.scss#L119

toolkit/scss/_search-result.scss
```diff
@@ -29,21 +39,18 @@
...
   background-color: $yellow-25;
   background-image: file-url('search-result-highlight-mask.png');
-  background-position: 0 -6px;
+  background-position: 0 -8px;
   background-repeat: repeat;
 }
 
 .search-result-metadata {
-  margin: 5px 0 0 0;
+  margin: 0 0 0 0;
```
tombye: This could do with reducing to a single `0`, like the padding style below.

quis: @tombye 149540a

toolkit/scss/_search-result.scss
quis: @tombye 149540a

PR #49, opened 2015-03-24
Add 'browse list' as a pattern

quis:
image


quis:
In reference to a conversation @tombye and I had in person: this pull request represents an iteration on the current homepage design which @ralph-hawkins has been working on. The intention would be to roll out this change from the toolkit to the Marketplace, whose homepage currently looks like this:

image

tombye:
:+1:

PR #51, opened 2015-03-26
Show code samples

quis:
This pull request:

- Replaces the `content` block in the YAML with one called `examples` which
  holds an array of code snippets
- Puts these code snippets onto the documentation pages in order
- Adds a `<pre>` block after each example for easily copy-and-pasteable code

## Examples

image

image

image


tombye:
It would be worth taking the `overflow: scroll` off the code examples. User research has shown they don;t mind having a long page if it means they can see all the code.

quis:
@tombye: a4326dc

quis:
@tombye: 2c44b8cbfc3610bc9667069b3135564aa39f77de

PR #52, opened 2015-03-26
Add option select GOV.UK component

quis:
What we have been calling "multi select" is now officially a GOV.UK component. This commit replaces the code from Finder Frontend with the component version from Static.

The point at which this is being brought in at is the merging of [the pull request](alphagov/static#556) I worked on recently.

This PR also renames "multi selects" to "option select" for consistency with the GOV.UK component.

Because of the markup and naming changes this PR represents a **breaking change**.


tombye:
:+1:

PR #53, opened 2015-03-26
Bump version to 2.0.0

quis:
Adds:
- #52 Add option select GOV.UK component 
- #51 Show code samples in documentation

**This release is a breaking change**. If you currently use the "multi select" pattern you will need to change:
- The names of the CSS and JS files you include in your project from `multi-select` to `option-select`
- The markup for your multi select controls


tombye:
Good notes :+1:

PR #62, opened 2015-04-30
Don't let elements that follow questions float

quis:
If a question contains floated elements then there is a risk that the following elements will float in an undesirable way. This change prevents that.

It also adds some spacing between questions to give them a bit more breathing room.

Example of the problem:

image

### DIFF
--- a/toolkit/scss/forms/_questions.scss
+++ b/toolkit/scss/forms/_questions.scss
@@ -3,9 +3,9 @@
 
 .question,
 %question {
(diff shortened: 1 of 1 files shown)


tombye:
Could we use `contain-floats` from the govuk front-end toolkit? It would mean we wouldn't need the `overflow: hidden` which might trip us up later at some point.

quis:
@tombye f977457

PR #65, opened 2015-05-08
Add search summary pattern

tombye:
The intention should be for the summary to always start with '**Count** services found in **lot**' and for the rest of the sentence to comprise of the filters applied, each with their own preposition and conjunctions.

For example, the following consists of the following filter sentences:

| filter name preposition | plural of the filter group name | options preposition | conjunction* | 
|--------------------|---------------------------------|----------------|-------------|
| under the          | categories                      |                | and         |
| with a             | minimum contract period         | of an          | or a        |

For filter options of more than 2, the conjunction listed would only feature at the end with all other options joined by commas.

search-summary


toolkit/scss/search/_search-summary.scss
```diff
@@ -0,0 +1,23 @@
...
+
+  @include media(tablet) {
+    @include core-16;
+  }
+
+  strong {
+    font-weight: bold;
+  }
+
+  .count {
```
quis: Could this be `.search-summary-count` to match our naming conventions?

tombye: Good point, fixed in edef4c1

PR #69, opened 2015-06-03
Add a pattern for contact details

quis:
image

image

This is adapted from the contact section of [a service page in the Grails app](https://www.digitalmarketplace.service.gov.uk/service/6574326446817280), but built in such a way that it could be used in other places, eg a supplier's dashboard.

It supplements the markup with microdata as specified by [schema.org/PostalAddress](https://schema.org/PostalAddress).

### DIFF
--- a/pages_builder/assets/scss/index.scss
+++ b/pages_builder/assets/scss/index.scss
@@ -12,6 +12,7 @@ $path : '../images/';
 @import "_page-headings.scss";
 @import "_search-result-wrapper.scss";
 @import "_browse_list.scss";
+@import "_contact-details.scss";
 @import "search/_search-summary.scss";
 @import "forms/_hint.scss";
 @import "forms/_list-entry.scss";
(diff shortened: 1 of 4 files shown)


TheDoubleK:
Was talking about contact details with @minglis the other day - vCard is the open government standard for contact details, so it would be good to have vCard format in the Marketplace.  This seems like a good opportunity to include it?

tombye:
Some useful reference for what @TheDoubleK is talking about:

https://www.gov.uk/government/publications/open-standards-for-government/exchange-of-contact-information

There's space in the HTML to include both the vCard and the microdata information though it does feel like we should just make a decision.

minglis:
vCard requires firstname/lastname and we only have name as per GDS form patterns. We have only inserted name as well into the schema. To do vCard we'd need to fix all the data. Will merge this and do a new story for vCard.

TheDoubleK:
Looks like vCard is blocked until this story is played: https://www.pivotaltracker.com/story/show/96271232

tombye:
The design does have the same semantics as vcard anyway so happy to merge it as is.

:+1:

PR #70, opened 2015-06-08
Add shared templates

quis:
## This documentation…
image

## …is now generated from this Jinja template
``` jinja
<div class="question">
  <label for="question-2" class="question-heading{% if hint is defined %}-with-hint{% endif %}">
    {{ question }}
  </label>
  {% if hint is defined %}
    <p class="hint">
      {{ hint }}
    </p>
  {% endif %}
  {% if large %}
    {% if max_length_in_words is defined %}
      <div class="word-count">
        <textarea class="text-box text-box-large" name="{{ id }}" id="{{ id }}" data-max-length-in-words="{{ max_length_in_words }}">{{ value }}</textarea>
      </div>
    {% else %}
      <textarea class="text-box text-box-large" name="{{ id }}" id="{{ id }}">{{ value }}</textarea>
    {% endif %}
  {% else %}
    <input type="text" name="{{ id }}" id="{{ id }}" class="text-box" value="{{ value }}" />
  {% endif %}
</div>
```

## And the templates can be used in other apps, like so:
image
(uses [Jinja's `with` extension](http://jinja.pocoo.org/docs/dev/extensions/#with-extension))

--
**The aim of this pull request is to allow sharing of templates between our apps**

It does this by:
- Add Jinja templates for all the design patterns in the toolkit
- Refactors the toolkit to generate the example markup from the Jinja templates and generate an example of the data passed to the template

This owes some debt to the work that @edds and @dsingleton did with [GOV.UK components](http://govuk-component-guide.herokuapp.com/).

This commit also does some renaming of the patterns to reflect exactly what they do. eg `search-results` (plural) is a template for rendering a list of search results, whereas `page-heading` (singular) is a template for rendering a single page heading. This commit only renames the examples and documentation, not the actual toolkit files (so it won't be a breaking change at this stage).


pages_builder/generate_pages.py
```diff
@@ -87,7 +94,41 @@ def render_page(self, input_file):
                 partial['pageTitle'] +
                 " - Digital Marketplace frontend toolkit"
             )
-            if "examples" in partial:
+            if (
```
tombye: Could we remove the brackets from this and line 105 to match the style adopted for the rest of this file?

pages_builder/generate_pages.py
```diff
@@ -87,7 +94,41 @@ def render_page(self, input_file):
...
+                "examples" in partial
+            ):
+                template_file = (
```
tombye: Be good to use [os.path.join](https://docs.python.org/2/library/os.path.html#os.path.join) to build this path.

pages_builder/generate_pages.py
```diff
@@ -87,7 +94,41 @@ def render_page(self, input_file):
...
+                )
+                if (os.path.isfile(template_file)):
+                    template = Template(
+                        open(template_file, "r").read()
+                    )
+                else:
+                    sys.exit("\n⚡ " + template_file + " not found ⚡")
+                for index, example in enumerate(partial["examples"]):
+                    parameters = dict(partial["examples"][index])
+                    if "title" in partial["examples"][index]:
```
tombye: Could we have a comment here explaining why `title` is removed? It's a bit unclear from just reading the code.

pages_builder/generate_pages.py
```diff
@@ -107,14 +148,18 @@ def render_page(self, input_file):
                                 {{#examples}}
                                     {{#title}}<h2>{{title}}</h2>{{/title}}
                                     {{{markup}}}
-                                    <pre>{{markup}}</pre>
+                                    <div class="code open"><p class="code-label">Template ({{ templateFile }})</p>{{{highlighted_markup}}}</div>
```
tombye: Looking at how they appear in the page, it would be good for these code labels to be `h2`s so the structure is:

h1: Name of the pattern
h2: Template
h2: Parameters

tombye:
Other than the comments above I've just got the following points:

I think the `id` parameter passed into patterns with radio buttons should be changed to `name`. It's used for that attribute in the HTML and appears in the HTTP parameters too. I think calling it `name` makes that usage more obvious.

The README should be updated to explain how the templates are part of the toolkit. Probably some updates to the [Example markup](https://github.com/alphagov/digitalmarketplace-frontend-toolkit/tree/add-shared-templates#example-markup) section and maybe a new section near [Sass compilation](https://github.com/alphagov/digitalmarketplace-frontend-toolkit/tree/add-shared-templates#sass-compilation) to explain where you need to work on HTML and where the template data is kept.

pages_builder/generate_pages.py
quis: Use of `os.path.join` and `os.path.splitext` addressed in f5269df

pages_builder/generate_pages.py
quis: Comment added in 684fb43

pages_builder/generate_pages.py
quis: Brackets gone as of 44d922e

quis:
@tombye Readme and naming of parameters to selection buttons added in 550de71 and 977ccc6 respectively.

pages_builder/generate_pages.py
quis: Heading structure added in f5269df85671f3e80588d4109933dc9930615be3

```
H1 – name of pattern
    H2 – description of example
        H3 – template
        H3 – parameters
```

tombye:
Managed not to say, this is brilliant work. Looking forward to pulling it down into the apps.

:+1:

dsingleton:
This is :ok_hand: :heavy_multiplication_x:  :100:

PR #74, opened 2015-06-10
Tweaks to templates

quis:
This does two things:
- Check for falseyness of error messages (rather than them being defined explicitly)
- Formalise street address as being two lines

This will make it easier to integrate the templates into the supplier app.

### DIFF
--- a/toolkit/templates/contact-details.html
+++ b/toolkit/templates/contact-details.html
@@ -1,9 +1,9 @@
-{% if organisation is defined %}
+{% if organisation %}
 <div class="contact-details" itemscope itemtype="http://schema.org/Organization">
   <p class="contact-details-organisation">
     <span itemprop="name">{{ organisation }}</span>
   </p>
-  {% if contact_name is defined %}
+  {% if contact_name %}
   <p class="contact-details-block">
     <span itemscope itemtype="http://schema.org/Person">
       <span itemprop="name">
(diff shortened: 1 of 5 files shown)


toolkit/templates/contact-details.html
```diff
@@ -35,10 +35,14 @@
...
+      {% if street_address.line_2 %}
+        <br />
+        {{ street_address.line_2 }}
+      {% endif %}
     </span>
     {% endif %}
-    {% if street_address is defined %}
+    {% if locality is defined %}
     <span itemprop="addressLocality">
       {{ locality }}
```
allait: Could we split `locality` into `city` and `country` too?

quis:
@allait Addressed in 46807ac (not calling it city because we should match the schema.org naming conventions as much as they make sense).

toolkit/templates/contact-details.html
```diff
@@ -28,28 +28,38 @@
...
+  locality or
+  country or
+  postcode
 %}
   <p class="contact-details-block{% if without_spacing %}-without-spacing{% endif %}" itemprop="address" itemscope itemtype="http://schema.org/PostalAddress">
-    {% if street_address is defined %}
+    {% if street_address %}
     <span itemprop="streetAddress">
-      {{ street_address }}
+      {{ street_address.line_1 }}
```
allait: Can we split `street_adress` into two variables instead? That would match two separate fields in our supplier data: `.address1` and `.address2`. I'm not sure if we can assign dictionaries in with statements to get around this, but I don't think we have an object with `.line_1` and `.line_2` anywhere so it might be more work anyway.

PR #82, opened 2015-06-17
Fix list entry

quis:
This commit adjusts how the parameters that are passed into the template are used, especially parameters used to identify elements.

Basically, it makes the `list-entry` pattern work in a real-world use case.

It ensures that:
- the `name` parameter of each textbox matches the key expected by the API
- there is a way of specifying a singular noun for what each item in a list is (eg a feature, a certification)
- there are no clashing IDs, even when the pattern is used multiple times on one page

### DIFF
--- a/pages_builder/pages/forms/list-entry.yml
+++ b/pages_builder/pages/forms/list-entry.yml
@@ -19,7 +19,7 @@ examples:
       - Three
   -
     question: With an error
-    id: service benefit
+    id: serviceBenefits
     error: Answer required
     number_of_items: 3
-    item_name: feature
+    item_name: benefit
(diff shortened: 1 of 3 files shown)


pages_builder/pages/forms/list-entry.yml
```diff
@@ -9,15 +9,17 @@ bodyEnd: >
...
     number_of_items: 10
+    item_name: service feature
     values:
       - One
       - Two
       - Three
   -
     question: With an error
-    id: feature
+    id: service benefit
```
tombye: Should this be camel-cased like `serviceFeatures`? It might actually be an idea to lowercase both and replace the space with a hypen instead as that's the pattern used elsewhere on these pages.

quis: @tombye Good point, addressed in 66afb1f69dad672bdff40dedb7e8a642754df841

PR #84, opened 2015-06-18
Update look and feel for validation pattern

quis:
This commit is built on top of the work and research that @gemmaleigh did towards
https://github.com/alphagov/govuk_elements/pull/71

It changes two main things:
- the borders and spacing of validation errors
- the nesting of markup to put the error message and hint inside the question's `label` or `legend`

It does not address:
- The language we use in the validation 'masthead', ie just referring users to which questions have errors, rather than explaining why the errors occurred (this is a more comprehensive change that would need to happen across all of our apps)

### Examples
---
screen shot 2015-06-18 at 16 05 29
---
screen shot 2015-06-18 at 16 05 22
---
screen shot 2015-06-18 at 16 06 15
---
image

### DIFF
--- a/toolkit/scss/forms/_validation.scss
+++ b/toolkit/scss/forms/_validation.scss
@@ -40,11 +40,6 @@
     margin: 0;
   }
 
-  .text-box {
-    border: 5px solid $red;
-    margin-bottom: 0;
-  }
-
 }
 
 .validation-message {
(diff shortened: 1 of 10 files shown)


toolkit/scss/forms/_validation.scss
```diff
@@ -3,8 +3,8 @@
 
 .validation-masthead {
 
-  border: 3px solid $red;
-  padding: 15px 20px 30px 20px;
+  border: 5px solid $red;
```
TheDoubleK: Because users might not take a 3px red line as an indicator of a problem?

PR #87, opened 2015-06-22
Add potential checked attribute to input elements.

pcraig3:
#### bref:
* Inputs now optionally accept a `checked` attribute.
* <del>Labels now optionally accept passed-in classes.</del>
* Top-level `inline` parameter will set `selection-button-inline` class on all labels.
* `<legend>` element given a `question-heading-with-hint` class if followed by a hint.

***
#### moins bref:

Was trying to convert this HTML into a couple of form elements.  (Presently in the [Supplier App](https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/master/app/templates/services/_edit-status.html#L4)).
```html+django
    <fieldset class="question">
      <legend class="question-heading question-heading-with-hint ">
        Choose service status
      </legend>
      <p class="hint">
        Private services don't appear in search results and don't have a
        <acronym title="Uniform resource locator">URL</acronym>
      </p>
      <label class="selection-button selection-button-boolean">
        <input type="radio" name="service_status" id="service_status_published" value="public" {% if service_data['status'] == 'published' %}checked="checked"{% endif %} />
        Public
      </label>
      <label class="selection-button  selection-button-boolean">
        <input type="radio" name="service_status" id="service_status_private" value="private" {% if service_data['status'] == 'enabled' %}checked="checked"{% endif %} />
        Private
      </label>
    </fieldset>
```

Needed to add checkboxes, a class to the legend, and I also needed that `selection-button-boolean` class in my labels, but I didn't want the rigidity that came with setting `type` to `boolean`.

<del>Added an `option.class` variable, which seemed like both an easy fix and a semantic one. 
I'm not crazy about it though -- it allows possibly undesirable flexibility to introduce new classes into the markup, and the way I'm using it right now it involves duplicating the classname in each of the inputs.</del>
Can toggle an inline class instead of passing in a classname.

I've only got two inputs, so it's working out alright for me, but if anyone has any better ideas I'm not averse to taking them.

### DIFF
--- a/toolkit/templates/forms/selection-buttons.html
+++ b/toolkit/templates/forms/selection-buttons.html
@@ -26,7 +26,7 @@
       </label>
     {% else %}
       {% for option in options %}
-        <label class="selection-button {{ option.class }}" for="{{ name }}-{{ loop.index }}">
+        <label class="selection-button{% if inline is defined %} selection-button-inline{% endif %}" for="{{ name }}-{{ loop.index }}">
           {{ option.label }}
           <input type="{{ type }}" name="{{ name }}" id="{{ name }}-{{ loop.index }}" value="{{ option.value or option.label }}" {% if option.checked %}checked="checked"{% endif %} />
         </label>
--- a/toolkit/templates/forms/selection-buttons.html
+++ b/toolkit/templates/forms/selection-buttons.html
@@ -26,7 +26,7 @@
       </label>
     {% else %}
       {% for option in options %}
-        <label class="selection-button{% if inline is defined %} selection-button-inline{% endif %}" for="{{ name }}-{{ loop.index }}">
+        <label class="selection-button{% if inline %} selection-button-inline{% endif %}" for="{{ name }}-{{ loop.index }}">
           {{ option.label }}
           <input type="{{ type }}" name="{{ name }}" id="{{ name }}-{{ loop.index }}" value="{{ option.value or option.label }}" {% if option.checked %}checked="checked"{% endif %} />
         </label>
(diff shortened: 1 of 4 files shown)


quis:
'Boolean' type fields are [rendered differently from normal radios/checkboxes](https://github.com/alphagov/digitalmarketplace-frontend-toolkit/blob/master/toolkit/templates/forms/selection-buttons.html#L16-L25).

We should think about how boolean ones can be pre set to yes/no as well.

pcraig3:
I know they are, but we want the inline style for the Public / Private status options.

So do we create a new class somewhere (`selection-button-inline`, maybe?), or do we let boolean radio buttons accept variables other than "Yes" and "No"?

quis:
I like `selection-button-inline` as well. Would be good to have it documented as part of [the pattern](http://alphagov.github.io/digitalmarketplace-frontend-toolkit/forms/selection-buttons.html) too.

pcraig3:
Added `selection-button-inline` class (which is toggleable), and updated the docs.

toolkit/templates/forms/selection-buttons.html
```diff
@@ -24,9 +26,9 @@
       </label>
     {% else %}
       {% for option in options %}
-        <label class="selection-button" for="{{ name }}-{{ loop.index }}">
+        <label class="selection-button{% if inline is defined %} selection-button-inline{% endif %}" for="{{ name }}-{{ loop.index }}">
```
quis: `if inline` would be better here. As is, `inline: False` does the opposite of what you'd expect.

pcraig3: Right, good catch.

PR #90, opened 2015-06-23
Add checked value for boolean selection button

robyoung:
If value is set to True yes is checked.
If value is set to False no is checked.


quis:
Would be good to update [the examples file](https://github.com/alphagov/digitalmarketplace-frontend-toolkit/blob/master/pages_builder/pages/forms/selection-buttons.yml) with a demonstration of the usage.

robyoung:
Good call, done.

quis:
Lovely stuff.

PR #94, opened 2015-06-30
Add textboxes for percentages and price

quis:
This commit adds two new patterns which will be useful in the SSP. The percentage one doesn't exactly match what was in the previous SSP (it had the '%' outset and in black), but it's more consistent to have it inset and in grey.

**A textbox with the unit after, eg for entering a percentage**

screen shot 2015-06-30 at 07 48 18

--

**A textbox with the unit before, eg for entering a price**

screen shot 2015-06-30 at 07 48 07

--

image

### DIFF
--- a/toolkit/templates/forms/textbox.html
+++ b/toolkit/templates/forms/textbox.html
@@ -5,6 +5,9 @@
     <label for="{{ name }}">
       <span class="question-heading{% if hint is defined %}-with-hint{% endif %}">
         {{ question }}
+        {% if unit_in_full %}
+          <span class="text-box-unit-qualifier">(in {{ unit_in_full }})</span>
+        {% endif %}
       </span>
       {% if hint is defined %}
         <span class="hint">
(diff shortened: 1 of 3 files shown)


toolkit/templates/forms/textbox.html
```diff
@@ -26,7 +26,19 @@
         <textarea class="text-box text-box-large" name="{{ name }}" id="{{ name }}">{{ value }}</textarea>
       {% endif %}
     {% else %}
-      <input type="text" name="{{ name }}" id="{{ name }}" class="text-box{% if error %}-with-error{% endif %}" value="{{ value }}" />
+      {% if unit == "before" %}
```
tombye: Could this be `unit_position` to match the pattern indicated by `unit_symbol`?

quis:
@tombye Had a think about it and changed the naming in c93686d

tombye:
That reads a lot nicer now, thanks!

tombye:
I did a bit of testing with Voiceover and the second label isn't getting read out. Would need more testing with JAWS but the solution may just be to put information about the unit into the label explicitly.

For example: 'Service price (in dollars)' or 'Percentage availability of the service'

quis:
@tombye I think in the SSP we had the information in the label and then used some absolute positioning jiggery pokery to get it looking right. Being explicit about it sounds sensible; changed in b13a8a1.

ralph-hawkins:
make-the-left-border-align

PR #96, opened 2015-06-30
Add pattern for previous next navigation

quis:
image

Currently this pattern is used in the buyer and admin apps but isn't in the toolkit.

This commit copies the GOV.UK component [from static](https://github.com/alphagov/static/blob/master/app/views/govuk_component/previous_and_next_navigation.raw.html.erb), as seen here:
http://govuk-component-guide.herokuapp.com/components/previous_and_next_navigation


toolkit/scss/_previous-next-navigation.scss
```diff
@@ -0,0 +1,127 @@
...
+// Media query helpers. These make producing IE layouts
+// super easy.
+
+// These are desktop and down media queries
+
+// There is also a local version of this in Smartanswers.
+
+$is-ie: false !default;
+
+@mixin media-down($size: false, $max-width: false, $min-width: false) {
```
tombye: This is legacy code from govuk which will, in time, be converted to mobile-up. Could we take this opportunity to make the switch here?

If we converted the `.govuk-previous-and-next-navigation` block to be mobile-up we also wouldn't need to include this mixin as we could just import `_conditionals.scss` from the govuk frontend toolkit.

quis: I'd argue for keeping it a verbatim copy-paste. Otherwise bringing in any updates to this pattern means remembering to and figuring out how to convert down to up.

Or, we make a PR against the component and fix it at the source…

tombye: Kind of annoying to include something that's following a different approach to the rest of our code but yes, I agree with the later point. I'll merge this and look to raise a PR against the component guide.

PR #99, opened 2015-07-06
Booleans should be true and false

TheDoubleK:
This allows booleans to save and display values.

It also allows checkboxes to be pre-checked for values that exist in the values passed in.

### DIFF
--- a/pages_builder/pages/forms/selection-buttons.yml
+++ b/pages_builder/pages/forms/selection-buttons.yml
@@ -25,16 +25,19 @@ examples:
     name: serviceStatus
     question: Inline radio buttons
     inline: True
+    value: 'Public'
     options:
       -
         label: Public
-        checked: true
       -
         label: Private
   -
     type: checkbox
     name: categories
     question: Choose a category
+    value: 
+      - Accounting and finance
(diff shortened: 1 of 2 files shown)


quis:
Going to need to update the docs since this changes the way selection-button is used. Can help you with this tomorrow.

TheDoubleK:
Docs updated.  I checked an inline radio but left the regular radios unchecked, and ticked a couple of checkboxes.

pages_builder/pages/forms/selection-buttons.yml
```diff
@@ -25,16 +25,17 @@ examples:
...
       -
         label: Public
-        checked: true
       -
         label: Private
   -
     type: checkbox
     name: categories
     question: Choose a category
+    value: ['Accounting and finance','Collaboration']
```
quis: Sorry for nitpicking—and I know this works—but it's more YAML-onic to do:
```YAML
value:
  - Accounting and finance
  - Collaboration
```

TheDoubleK: No worries, yep it doesn't look quite right now you say it.

PR #100, opened 2015-07-07
Bump version to 6.0.0

quis:
Includes
- #99 Booleans should be true and false

**MAJOR** change because it alters the parameters passed to selection buttons. If you have been passing a `checked` attribute to any `options` then this should be removed and replaced with a string/array of current values for the whole set of options.

🍰_Pull request_ 💯


TheDoubleK:
Where is the cake?

PR #101, opened 2015-07-08
Draw custom underlines for links in browse-list

quis:
image

--

We have noticed in user research and from feedback that support has received that it is not always obvious that these are links.

Adding an underline to these links should help, but the default underlines on large text are heavy and make the text less legible (or that's what we intuit anyway).

The GOV.UK template uses a CSS `border` for drawing a nice, 1px underline on the GOV.UK logo when hovered. So there is a visual precedent.

Using borders doesn't usually work so well on text, because they sit a lot lower than the normal position of an underline. ~~Using box-shadow instead allows control of the vertical position of the underline, as well as the thickness (thickness is managed by drawing two shadows, one for the blue and a white one to mask it).~~  This can be worked around with a  `<span>` nested inside the link in order to relatively position
the text at a vertical offset from the bottom border.

--

image
<sub>Latest Chrome, OS X Yosemite</sub>

--

image
<sub>Samsung Galaxy S4, Android 4.4</sub>

--

For older browsers we fall back to a normal `text-decoration: underline` (IE9 and below) or no underline (_really_ old non-IE browsers that don't support box-shadow).

image
<sub>IE 8, Windows XP</sub>

image
<sub>Firefox 3.6, Windows XP</sub>

-- 

This pull request also makes slight changes to the spacing and type size of the text underneath the links.

### DIFF
--- a/toolkit/scss/_browse_list.scss
+++ b/toolkit/scss/_browse_list.scss
@@ -2,6 +2,7 @@
 @import "_typography.scss";
 @import "_measurements.scss";
 @import "_shims.scss";
+@import "_conditionals.scss";
 
 .browse-list {
   padding: 0;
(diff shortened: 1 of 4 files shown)


quis:
Just discovered this problem with IE11. Working on a fix.

image

quis:
IE11 fix in 81377c8:

image

toolkit/scss/_browse_list.scss
```diff
@@ -12,20 +13,76 @@
...
-  @include inline-block;
-  @include heading-27;
+.browse-list-item-link,
+a.browse-list-item-link {
+
+  display: inline;
+  @include heading-24;
   font-weight: bold;
   text-decoration: none;
+  border-bottom: 1px solid;
```
tombye: Would be nice to add a comment to explain why there's no colour specified here so it's clear it wasn't missed.

tombye:
Apart from my comment, LGMT :+1:

toolkit/scss/_browse_list.scss
quis: @tombye Comment added in 70731f1

PR #107, opened 2015-07-16
Refactor summary pattern

quis:
This commit makes it possible to:
- have arbitrary numbers of columns
- specify whether column headings should be shown (and styled) or hidden
- use the template, rather than having to copy/paste the markup

It also alters some type sizes and weights to better match the GOV.UK pattern and make the top-level links more obvious.

Full documentation
--
summary - digital marketplace frontend toolkit

Jinja code for the last example
--
``` Jinja
{%
  with
  rows = [
    {
        "fields": [
            "Summary item name", 
            "<a href=\"#\">Summary item content</a>"
        ], 
        "incomplete": true
    }
  ],
  top_link = "#",
  heading = "Summary item with incomplete entry",
  top_link_label = "Edit <span class=\"visuallyhidden\">summary item with incomplete entry</span>"
%}
  {% include "toolkit/templates/forms/summary.html" %}
{% endwith %}
```

### DIFF
--- a/toolkit/scss/forms/_summary.scss
+++ b/toolkit/scss/forms/_summary.scss
@@ -33,10 +33,22 @@
   padding-left: 75%;
   position: relative;
   z-index: 10;
+  font-weight: bold;
 
   @include ie(6) {
     zoom: 1;
   }
+}
+
+.summary-item-no-content {
+  color: $secondary-text-colour;
+  @include core-16;
+  margin: 0 0 15px 0;
+  border-top: 1px solid $grey-3;
+  border-bottom: 1px solid $grey-3;
+  padding-top: 10px;
+  padding-bottom: 10px;
+  max-width: 100%;
 }
 
 %summary-item-body {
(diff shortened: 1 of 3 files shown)


toolkit/scss/forms/_summary.scss
```diff
@@ -33,71 +33,86 @@
   padding-left: 75%;
   position: relative;
   z-index: 10;
+  font-weight: bold;
```
tombye: If the `copy-19` above is changed to `bold-19` we won't need this line. Would be worth doing I think.

toolkit/scss/forms/_summary.scss
```diff
@@ -109,27 +124,38 @@ table.summary-item-body {
 
 }
 
-.summary-item-field-content {
-  @extend %summary-item-field;
-  width: 66.666%;
-  padding-left: 25px;
+.summary-item-field-headings-visible {
+  @include core-16;
+  font-weight: bold;
```
tombye: Same comment as above here, it would be good to swap out the `core-16` for `bold-16` to save this line.

tombye:
Apart from my comments, LGTM :+1:

toolkit/scss/forms/_summary.scss
quis: It won't work here because `bold-19` is not a drop-in replacement for `copy-19` (different padding). Have changed it where possible in 0716073.

PR #109, opened 2015-07-20
Allow HTML to be passed to summary table cells

quis:
Table cells can contain things like:
- lists
- [links](#)

…and we've decided that the prefered way to do this is build the HTML in the app and pass it through to the toolkit, like we do with the beta banner, etc.

The alternative would be to have the pattern know about things like assurance approaches, which would make our toolkit very tightly coupled to our current, limited set of data structures.


pcraig3:
Looks `|safe` to me.

PR #111, opened 2015-07-21
Pass HTML to table summary headings, make table heading optional.

pcraig3:
##### Fix `onclick` behaviour for code examples.

Code examples open when clicked, but would also collapse when
any part of them were clicked.  This meant it was difficult to
practice good CPDD<sup>1</sup>.

##### Allow HTML to be passed to summary table headers

Allow passed-in HTML markup to render in table headers.
This means we can give classes or links to individual table headers.
Same as we did with table cells in a past commit (914d401)

##### Make table heading optional

If no heading is given, no heading appears in the resulting HTML.
Previously, an empty `h2` tag would still be generated.

<sub>1: Copy-Paste Driven Development</sub>

### DIFF
--- a/pages_builder/assets/javascripts/onready.js
+++ b/pages_builder/assets/javascripts/onready.js
@@ -27,8 +27,8 @@
     }
   }
 
-  $('.code').removeClass("open").click(function() {
-      $(this).toggleClass('open');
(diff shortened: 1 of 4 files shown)


quis:
:+1:

pcraig3:
:rocket:

PR #113, opened 2015-07-22
Fix hidden table headings

tombye:
The previous method of hiding the column headers was causing their containing row to be ignored by accessibility APIs:

html_node_accessibility_info

It also meant each table was being identified as two: the hidden row and the rest of the table.

voiceover_table_information

This fixes that issue and adds captions to all tables meaning they get an accessible label which can be used to identify them in isolation from their place in the document.

voiceover_table_information_with_captions

### DIFF
--- a/pages_builder/pages/forms/summary.yml
+++ b/pages_builder/pages/forms/summary.yml
@@ -2,6 +2,7 @@ pageTitle: Summary
 assetPath: ../govuk_template/assets/
 examples:
   -
+    caption: Summary item with no entries
     field_headings:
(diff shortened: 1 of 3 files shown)


quis:
Passing the caption as a separate parameter means there's no problem with tables without a `<h2>` above them.

tombye:
Looks like it's also part of the HTML5 spec to include it: http://www.w3.org/TR/html5/tabular-data.html#the-caption-element

quis:
Had a look locally, :+1:

PR #116, opened 2015-07-23
Macro-based summary pattern template

allait:
### Rewrite forms/summary pattern using Jinja2 macros
Splits the summary pattern into separate macros: for summary
header, top action link and the summary table.

Summary table relies on the Jinja2 "macro call" syntax
(http://jinja.pocoo.org/docs/dev/templates/#call).

This allows use of the pattern to create the table layout while
row contents are still written inside the template HTML.

Helper macros are provided for common field types.

This avoids the need for `| safe` in the pattern template.

### Update pages builder script to support Jinja2 snippets as examples
Adds a check for the type of "example": if example is a dictionary
then `with` parameters example is generated, otherwise example is
rendered as a Jinja template.

Adds Environment to the example templates, which allows to use includes
and imports relative to the `toolkit/templates` folder.


quis:
Other than that one comment, colon plus symbol one colon from me.

This is a really neat solution to a difficult problem.

robyoung:
I found the link to the macro documentation really helpful in understanding this. It would be good if it were included in the docs somewhere.

allait:
Added assurance argument and a `pageDescription` field to the documentation files to link to the Jinja2 macro call docs /cc @quis @robyoung

robyoung:
plus one from me

PR #121, opened 2015-07-28
Allow text field macro to accept assurance

quis:
Currently calling the text field macro like this:

```Jinja
  {{ summary[item.type](item.value, assurance) }}
```

results in an error, because the text field only accepts one parameter. This means handling it as a special case.

It would be easier if it had the same interface as `list`, and any other macros that might follow (eg document link)


allait:
:+1:

PR #123, opened 2015-07-29
Add a Makefile for generating and serving pages

robyoung:
This just makes it a bit easier to generate and serve pages on the toolkit.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,13 @@
+SHELL := /bin/bash
+
+virtualenv:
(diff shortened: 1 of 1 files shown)


robyoung:
Talked to @chs in meatspace and he's ok with it.

robyoung:
Testing testing...
```jinja
{% if true %}
<h1>YAY</h1>
{% endif %}
```

quis:
:+1: in non-meatspace too

PR #124, opened 2015-07-29
Add pricing form

robyoung:
Add a template for the pricing form. The markup and SCSS is mostly taken
from the [admin app](alphagov/digitalmarketplace-admin-frontend@86afb5a). This differs from other form templates in that
it is made up of multiple form elements and therefore requires multiple
default values. This is expected to be handled in macros in the
frontends. eg.

```jinja
{% macro pricing(question_content, service_data, errors) -%}
  {%
    with
    name=question_content.id,
    priceMin=service_data.priceMin,
    priceMax=service_data.priceMax,
    priceUnit=service_data.priceUnit,
    priceInterval=service_data.priceInterval,
    hint=question_content.hint,
    error=errors.get(question_content.id)['message']
  %}
    {% include "toolkit/forms/textbox.html" %}
  {% endwith %}
{%- endmacro %}
```

screen shot 2015-07-29 at 16 41 17

screen shot 2015-07-29 at 16 41 28

### DIFF
--- a/pages_builder/assets/scss/index.scss
+++ b/pages_builder/assets/scss/index.scss
@@ -19,6 +19,7 @@ $path : '../images/';
 @import "forms/_hint.scss";
 @import "forms/_list-entry.scss";
 @import "forms/_option-select.scss";
+@import "forms/_pricing.scss";
 @import "forms/_questions.scss";
 @import "forms/_selection-buttons.scss";
 @import "forms/_summary.scss";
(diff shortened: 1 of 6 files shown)


quis:
:+1: 

I you want to close https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/125 and fold all of these into one version I would be OK with that.

PR #126, opened 2015-07-30
Do falsy checking for hint in forms

quis:
Checking for `is defined` means that even if a question's hint is `""` or `None` the hint's containing element will still be rendered, possibly with the word 'None' inside it.

This commit changes the toolkit templates to be less strict and only check for the falsyness of question hints.


allait:
:+1:

PR #127, opened 2015-07-31
Allow for 'radio' questions to be lists or strings

quis:
Questions which the SSP collected using radio buttons are sometimes returned by the API as strings and sometimes as lists.

This commit adds a check to the 'radios' macro to avoid accidentally passing a string to the 'list' macro and throwing an error.


allait:
:+1:

PR #128, opened 2015-07-31
Wrap summary-table comments with Jinja comment syntax

allait:
If the template is included instead of imported it would insert any
text into the template html. Wrapping strings with Jinja comments
makes sure they'll never appear in the HTML output.


quis:
:+1:

PR #130, opened 2015-08-04
Improve pricing form module

robyoung:
- Add a bit of space between the main hint and the field hints
- Make the word Optional in the field hints grey
- Fix typo in visually-hidden class name

## Before
screen shot 2015-08-04 at 11 47 11

## After
screen shot 2015-08-04 at 11 46 38


quis:
:+1:

PR #132, opened 2015-08-05
Use appropriate checks for previous values

quis:
The `value` of checkboxes come back as an array, so the appropriate check is to look for any matching value in the array.

The `value` of radio buttons come back as a string, so it is more appropriate to do a simple equality check.


allait:
:+1:

PR #133, opened 2015-08-05
Fix summary-table upload field display

allait:
Since presenters aren't used for the summary-table the macro value
argument is the document URL instead of a `{url, filename}` dict.

The macro not does the same filename processing that the upload
presenter used to do.


quis:
Could you add a row containing a URL to the 'Summary table with various types of service content' example in the docs?

allait:
@quis done

quis:
:+1:

PR #135, opened 2015-08-06
Fix bug with selection buttons

quis:
`radios` and `checkboxes` are **not** valid parameters for the `type` of this template.

`radio` and `checkbox` are.

This is currently breaking all selection buttons if you use the latest toolkit version.


allait:
:+1:

PR #137, opened 2015-08-06
Textboxes should be aware of None when rendering values

minglis:
- Passing a WTF form into a textbox can result in None being rendered as value
- Checking explicitly for None to ensure this doesn't happen


quis:
You can't have whitespace inside a `<textarea>`:

image

quis:
:+1:

PR #138, opened 2015-08-10
Add link button pattern

quis:
This is a pattern we use in the admin app to link off to the view of a service in the buyer app.

It would be good to have it shared, so that it could be used the same way in the supplier app.

image


ralph-hawkins:
I have some reservations about the button-like appearance and the positioning makes it feel a bit disconnected from the context of the service. But, happy for it to go live for consistency between the two apps, we can fix both at some point.

In summary... LGTM :+1: 😔

PR #139, opened 2015-08-12
Hide empty breadcrumbs

allait:
### Don't show breadcrumb elements with missing labels
Only shows the breadcrumb element if it has a non-empty label value.
Since links and text with empty values are not visible anyway, this
hides the trailing ">" markers.

Both valid link and text breadcrumbs should have labels, but setting
label to an empty value is an easy way to filter the breadcrumbs on
some pages.


quis:
I have no idea what magic you are doing that requires this functionality, but looks reasonable to me :+1:

allait:
@quis hiding the "New service" breadcrumb for unnamed services, so that there's no way to go from that page to edit other sections without ever entering a name (since unnamed services won't be shown in the drafts list).

PR #140, opened 2015-08-13
Add question numbers

quis:
These are currently used in the supplier declaration, but the semantics aren't great because there's no way of injecting them into the label/legend of a field template in the toolkit.

This commit adds the option to pass in a number to each field template.

image


allait:
:+1:

PR #141, opened 2015-08-13
Use Javascript to jump from error summary to field

quis:
~~This commit changes the link in the validation masthead to point at the whole question, not just the field within it.~~

~~There is not always a one-to-one relationship between error and field, eg where a question has a series of checkboxes. There is always a one-to-one relationship between summary error message and question wrapper.~~

This follows the recommendations of GOV.UK elements, which can be seen here:
http://govuk-elements.herokuapp.com/examples/form-validation-multiple-questions


tombye:
Is there a test page for this so it's possible to see how this would work with keyboard usage and screenreaders?

The reason for linking to a form control was usually because it's focusable. Not sure how the question divs would work as focusable elements so I'd be keen to test this.

quis:
You can test with any of the documentation pages for forms—textbox, selection buttons, etc. They all have example of what happens when there's an error.

tombye:
I can't see a validation masthead on any of them. Any help?

quis:
Rewrote this in adfd584 to follow the way it's done [in GOV.UK elements](http://govuk-elements.herokuapp.com/examples/form-validation-multiple-questions).

allait:
:+1:

PR #143, opened 2015-08-25
Fix field summary heading style

robyoung:
summary-item-column-heading does not exist. If the first field was the only visible one then the height would not be set.

screen shot 2015-08-25 at 09 40 22


quis:
:+1:

PR #145, opened 2015-08-27
More aggressive banner style...

ralph-hawkins:
that also looks nicer, not tested. 

Is more consistent with the banners on things were you fill out question.

Old:

before

New:

after


quis:
Slightly different green:

image

PR #146, opened 2015-08-27
Make buttons work in old IE, and in combination

quis:
This pull request:
- adds `type='submit'` so that they will submit their parent form in all browsers (without this they do nothing in IE 7 and below)
- adds an optional `name` parameter which gets submitted with the other elements of a form, so that the backend can detect which button has been pressed
- changes from a `button` element to an `input` element so that the above works consistently in all browsers


allait:
9.0.0 :+1:

PR #147, opened 2015-08-27
Optionally put hint underneath selection buttons

quis:
image


tombye:
I wonder (and this is just that, not a solid reckon) if we should be adding position to the hint parameter and making that of the language we use to describe questions? 

Like: `hint-below-question` and `hint-below-answers`

...rather than

`hint` and `hint-below-answers`

(Like the rules we use for CSS class names.)

PR #150, opened 2015-09-03
[#102562916] Match selection buttons on value or label

TheDoubleK:
This partially fixes this bug: https://www.pivotaltracker.com/story/show/102562916

There is another part to the bug which needs to be fixed in supplier frontend.

The assurance questions now have different values (Uppercased) from labels (all lower), so the buttons were not being checked where previous values exist.

This change prefers `option.value` to `option.label` to match on, but will continue to match on `label` if no separate value is specified.


quis:
:+1:

PR #151, opened 2015-09-04
[#102562916] Allow none values for checkboxes

TheDoubleK:
This is required as part of this bugfix: https://www.pivotaltracker.com/story/show/102562916


quis:
:+1:

PR #152, opened 2015-09-07
Rename _browse_list.scss to _browse-list.scss

quis:
Because consistency.

--

Breaking change because any apps using the browse list pattern will have to update their `application.scss` file.


tombye:
:+1:

PR #153, opened 2015-09-11
Remove extraneous slash in Jinja partial examples

quis:
Something that wasn't caught doing https://www.pivotaltracker.com/story/show/100557872

--

image

image


tombye:
:+1:

PR #155, opened 2015-09-16
Use bold for search highlights

quis:
Matches the way that it's now done on GOV.UK: alphagov/frontend#879

image


tombye:
:+1:

rboulton:
Nice work to keep things consistent. :+1:

PR #156, opened 2015-09-17
Ensure highlights have no background colour

quis:
image

Highlights were changed to use the <mark> element in alphagov/digitalmarketplace-search-api#54

Some browsers, by default, style the mark element with a bright yellow background.

This makes sure that doesn't happen.


tombye:
:+1:

PR #158, opened 2015-09-18
Add grids to form content

tombye:
This includes changes to the CSS for the forms patterns so their horizontal spacing is controlled by the grid instead of being defined in their CSS. One of these grids has also been added to each of the form patterns example pages to show the expected use.

This work includes changes to the code that generates the pages to allow them to have grids to wrap the pattern. You now just need to add a `grid` parameter containing the class name of the column type, for example:

https://github.com/alphagov/digitalmarketplace-frontend-toolkit/blob/add-grids-to-form-content/pages_builder/pages/forms/textbox.yml#L3

The input list and pricing patterns have both had additional changes made to their CSS to allow them to work when sitting inside a two-thirds grid column:

input_list

pricing

The pricing pattern has also been made to work for mobile screens which involved some consultation with @quis:

pricing_mobile_view

### DIFF
--- a/pages_builder/generate_pages.py
+++ b/pages_builder/generate_pages.py
@@ -107,6 +107,11 @@ class Styleguide_publisher(object):
             }
         )
 
+    def render_include(self, include_file_name, data):
+        template_file = os.path.join(self.repo_root, "pages_builder", "includes", include_file_name)
+        template = open(template_file, "r").read()
+        return pystache.render(template, data)
+
     def render_page(self, root, file):
         input_file = os.path.join(root, file)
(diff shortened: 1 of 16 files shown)


quis:
image

quis:
All looks good apart from the bug above.

We talked in person about making textboxes appropriately wide for the kind of content they're meant to hold, but agreed that that's a separate pull request/story.

tombye:
@quis good spot with the word count: I'd done that work but neglected to add it. Now added in c0066d4

quis:
:+1:

PR #159, opened 2015-09-18
Make JavaScript tests run in CI

tombye:
Uses the same testrunner as in other apps.


quis:
I think you need to add a `travis.yml` like [the one in the supplier app](https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/master/.travis.yml) to actually get the tests to run.

allait:
I've made a change in Travis CI to enable running tests on this repo, but I think it needs a git push to trigger it for the first time

quis:
Pushed an empty commit to this branch, tests are running here: https://travis-ci.org/alphagov/digitalmarketplace-frontend-toolkit/builds/81363696

…and they've passed, but haven't run any of the tests.

Going to reset this branch to before I meddled now.

allait:
Travis shows a "This job ran on our legacy infrastructure" warning on these builds, even though recently created projects should default to the container builds. I think we should add `sudo: false` to `.travis.yml` to fix this

tombye:
OK, so the problem was the specs were listed in `spec/javascripts/manifest.js` with a lowercase letter to start but the files started with an uppercase letter. 

Looks like this was working on our Macs because our hard drives are formatted to have case-insensitive file handling.

quis:
Good sleuthing :+1:

PR #160, opened 2015-09-21
Bump version to 11.0.0

tombye:
Brings in [#158](https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/158) which makes all form content rely on grids for its width.

This makes it a breaking change for all pages using these patterns as they will have to wrap the content of their forms in a grid for them to display correctly.


quis:
:+1:

PR #161, opened 2015-09-22
Make grids match govuk toolkit examples

tombye:
The supplier and buyer apps both use the class names suggested by the govuk front-end toolkit rather than those used on our [grids page](http://alphagov.github.io/digitalmarketplace-frontend-toolkit/grids.html):

- [buyers app grids scss](https://github.com/alphagov/digitalmarketplace-buyer-frontend/blob/master/app/assets/scss/_grids.scss)
- [suppliers app grids scss](https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/master/app/assets/scss/_grids.scss)

It's important that we use the same classes across all the apps because some of the styles that make the form content work inside grids are tied to them (see [this example](https://github.com/alphagov/digitalmarketplace-frontend-toolkit/blob/master/toolkit/scss/forms/_pricing.scss#L11)).

I propose we use the class names most in use across our stack, this switches to them in the toolkit.

### DIFF
--- a/pages_builder/assets/scss/grids/index.scss
+++ b/pages_builder/assets/scss/grids/index.scss
@@ -3,27 +3,6 @@
 @import "_colours.scss";
 @import "_css3.scss";
 
-.grid-row {
-  @extend %grid-row;
-  @extend %outdent-to-full-width;
-
-  .third-column {
-    @include grid-column(1/3);
-  }
-  .two-third-column {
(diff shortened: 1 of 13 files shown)


quis:
:+1: 

1280px-bos_grunniens_at_letdar_on_annapurna_circuit

PR #162, opened 2015-09-23
Bump version to 11.0.1

tombye:
Includes making all grids class names use the examples from the govuk front-end toolkit.

See https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/161


quis:
https://github.com/alphagov/digitalmarketplace-frontend-toolkit/compare/v11.0.0...bump-to-11.0.1

Think this *is* a minor bump because none of the apps are using the existing `grids/index.scss` file.

:+1:

PR #163, opened 2015-09-24
Add option select search box

tombye:
Adds the text filter we use on the buyer and admin apps.

keyword_search_pic

Also can be seen on preview at https://preview.development.digitalmarketplace.service.gov.uk/g-cloud/search?lot=saas

### DIFF
--- a/toolkit/scss/forms/_option-select.scss
+++ b/toolkit/scss/forms/_option-select.scss
@@ -132,6 +132,8 @@
   }
 }
 
+// Styles specific to Digital Marketplace
+
 .govuk-option-select {
   margin-bottom: $gutter;
 
--- a/toolkit/scss/forms/_option-select.scss
+++ b/toolkit/scss/forms/_option-select.scss
@@ -131,3 +131,15 @@
     }
   }
 }
+
+.govuk-option-select {
+  margin-bottom: $gutter;
+
+  .container-head .option-select-label {
+    display: block;
+  }
+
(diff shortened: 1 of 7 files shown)


toolkit/scss/forms/_option-select.scss
```diff
@@ -131,3 +131,15 @@
     }
   }
 }
+
+.govuk-option-select {
```
quis: Putting this at the bottom of the file means it could be very easily overwritten if we ever copy/pasted in changes from the GOV.UK component. Maybe put it above the comment about the origin of the code in this file?

tombye: I agree about the danger but putting it above the copied block seems a bit odd seeing as it being below means the cascade reflects the order of inheritance:

GOVUK option-select component -> Digital Marketplace option-select (in DM Toolkit) -> app-level styles

What about putting a comment above this block to make it clear the separation instead?

quis: Yeah, or split it into two files which this file then includes? 

I'll thumbs up either of the above.

quis:
:+1:

tombye:
@quis I've added a comment to the DM-specific styles in `_option-select.scss` to identify them as such. I did consider splitting them and the styles taken from GOVUK components out into separate files but, given the name of the parent file, I thought there could be issues with naming that would confuse understanding of the CSS.

PR #164, opened 2015-09-24
Add JavaScript for module loading

tombye:
Runs through every object or method of the `GOVUK.GDM` namespace and initialises them.

This is based on the following assumptions:

1. if the child is an object it will be initialised by running its `init` method
2. if the child is a method, running it will initialise it

This also contains stubs for the `console` object to protect against errors in browsers which do not support it.

It also has awesome graphics (in your console):

image


quis:
:+1:

PR #166, opened 2015-09-28
Format service IDs

robyoung:
Format service IDs

This change splits service IDs for display if they are all digits. This functionality was previously implemented in the applications.


allait:
Can we implement the filter as a Jinja macro to avoid the dependency on dmutils?

robyoung:
Yes it can.

quis:
Neat :+1:

PR #168, opened 2015-09-29
Fixes to scripts

quis:
## Use git describe to find previous version

`git tag` returns tags alphabetically, not by date.

`git describe` is the appropriate command, thanks Stack Overflow: http://stackoverflow.com/a/14337129/147318

## Remove redundant line
-1


tombye:
:+1:

PR #169, opened 2015-10-02
Add "temporary message" pattern

quis:
This is used currently for notifying suppliers of the status of a framework,
and on the Digital Services Store as a warning about unsigned framework
agreements.


TheDoubleK:
Looks good :+1: 
Does the version number bump go into a separate PR?

PR #170, opened 2015-10-02
Fix tagging of versions

quis:
- get the previous version by looking at the most recent tag to head before creating the new tag
- show changes up to head instead of up to version bump, because a merge happens after a version bump commit if they're bundled in one PR


TheDoubleK:
YOLO - let's give it a go :+1:

PR #171, opened 2015-10-06
Add padding to temporary message lines

TheDoubleK:
Ralph likes it like this: http://dm-prototype.herokuapp.com/states/homepage?g7.framework=inReview

Before:
screen shot 2015-10-06 at 10 46 32

After:
screen shot 2015-10-06 at 10 46 16

@tombye - I'm not sure what impact (if any) this change will have on the supplier app messaging you've been working on?


tombye:
Paragraphs in temporary messages already have 5px bottom margin so I'm not sure the first screenshot is an accurate view of what they looked like before this change (it looks like there is no vertical spacing between them).

I'd also say that, if we are going to add more spacing, we should just increase the existing `margin-bottom` (I know this would ruin the naming of the branch :S). Oh and that it would be nice to remove the spacing for the last paragraph. Something like:

```
.temporary-message-message:last-child {
  margin-bottom: 0;
  padding-bottom: 0;
}

tombye:
Oh yeah, I added a commit to put the page for this into the github pages index as I saw it was missing.

PR #172, opened 2015-10-08
Fix option select content visibility

tombye:
Brings in bug fixes for option-select, specifically one which sets the options container to `display: none` when it is hidden, removing it from the document flow.

This counters the undesirable behaviour Léonie picked up on whereby an option-select would open when you tabbed to it. This meant you couldn't tab through the page without having to go through each option in each option-select.

Mentioned by @ralph-hawkins.


tombye:
Talking with @robyoung I realise our usage is different to that on www.gov.uk.

They fire searches for each option clicked (using AJAX) so it doesn't matter that some options could be out of the document flow. 

This isn't the case for us. Our searches are submitted by a normal form submission so options hidden in this way won't be posted as HTTP parameters. This will need a fix first so closing for now.

PR #173, opened 2015-10-22
[#105005304] Add framework-notice template (in addition to temporary-message)

pcraig3:
There are (at this point) two types of framework notice messages we need:

- The blue one [we currently have](https://www.digitalmarketplace.service.gov.uk/).
screen shot 2015-10-22 at 11 46 12

- The more descriptive one that [we're introducing DOS with](http://dm-prototype.herokuapp.com/states/homepage?dos.framework=open).
screen shot 2015-10-22 at 11 38 40

The first one [we already have](http://alphagov.github.io/digitalmarketplace-frontend-toolkit/temporary-message.html), and the latter one is added in this pull request. 

To get the flexible formatting we need, I've included a markdown filter in the template, which will be [included in digitalmarketplace-utils](https://github.com/alphagov/digitalmarketplace-utils/blob/master/dmutils/flask_init.py#L48) and subsequently available to all of our apps.  The [`markdown_filter`](https://github.com/alphagov/digitalmarketplace-utils/blob/master/dmutils/filters.py) method to utils also allows us to add it to our jinja environment in `generate_pages.py` so that our pages build. 

Should work everywhere with minimal effort. :smile:

Related to [this story](https://www.pivotaltracker.com/story/show/105005304) on Pivotal.

### DIFF
--- a/toolkit/templates/framework-notice.html
+++ b/toolkit/templates/framework-notice.html
@@ -8,6 +8,6 @@
     </p>
   {% endif %}
   {% if message %}
-    {{ message }}
+    {{ message|markdown }}
   {% endif %}
 </aside>
--- a/toolkit/templates/framework-notice.html
+++ b/toolkit/templates/framework-notice.html
@@ -1,9 +1,9 @@
 <aside role="complementary" aria-labelledby="framework-notice-heading" class="framework-notice">
-  <h2 id="framework-notice-heading">
+  <h2 class="framework-notice-heading">
     {{ heading }}
   </h2>
   {% if subheading %}
-    <p class="framework-notice-subheading hint">
+    <p class="framework-notice-subheading">
       {{ subheading }}
     </p>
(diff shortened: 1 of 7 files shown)


toolkit/templates/framework-notice.html
```diff
@@ -0,0 +1,13 @@
+<aside role="complementary" aria-labelledby="framework-notice-heading" class="framework-notice">
+  <h2 id="framework-notice-heading">
+    {{ heading }}
+  </h2>
+  {% if subheading %}
+    <p class="framework-notice-subheading hint">
```
quis: Can this just have one class and put `colour: $secondary-text-color` in the SASS instead?

pcraig3: `color: $secondary-text-colour;` 

:+1:

quis:
:+1:

PR #175, opened 2015-10-22
Add missing import

quis:
This was causing the Jenkins build of the documentation to fail


pcraig3:
Darn, I had fixed that and then rebased it out somewhere.

pcraig3:
:+1:

PR #178, opened 2015-11-09
Add statuses to browse list pattern

quis:
The browse list pattern needs to be extended to show the count of services in each lot.

This commit adds a 'status' block, which takes an array of messages, each of which can optionally be important (bold).

Follows the prototype at http://dm-prototype.herokuapp.com/states/services-dos?dos.application=all
…except for the width of the border, which I’ve kept at 5px to match the borders on the validation error pattern.

--

image


TheDoubleK:
Looks pretty :+1:

ralph-hawkins:
:+1: LGTM

PR #181, opened 2015-11-17
Improve documentation navigation

quis:
This pull request:
- sorts the navigation in the documentation alphabetically
- flattens the navigation to add links to form patterns on the homepage

image


TheDoubleK:
In the picture the nested form elements aren't in alphabetical order but in the code they are. Assuming it's an old screenshot and they do now display in order then :+1:

(Can't easily test on my phone, but the code looks right)

PR #182, opened 2015-11-17
Preserve key order when parsing YAML

quis:
Reasons:
- it makes it easier to cross-reference the code examples and the YAML files and the outputted code examples
- the YAML files have their keys in a logical order, eg `title` comes before `body`; it’s good to keep this in our code examples, and, by extension, the code that gets copied and pasted into our apps

## Before
image

## After
image


TheDoubleK:
Nice one! :+1:

PR #184, opened 2015-11-18
Add multiquestion macro.  

pcraig3:
A `multiquestion` macro allows us to display multiple answers in one summary table column cell. (See screenshot).

To achieve this, it was necessary to separate the `field()` calls (which create a new table cell) from the markup containing the content. This lets multiquestions call upon the content-only macros without creating a whole whack of new table cells.

Also created a new file, `summary-content.html` with the markup for presenting the service data, and retrofitted the original macros to call upon the content-only versions.

***

### Rendering a Multiquestion

screen shot 2015-11-16 at 11 25 14


quis:
This seems like a sensible approach, but it also seems to have broken all the things:

image

pcraig3:
?? wat

pages_builder/pages/summary-table.yml
```diff
@@ -224,6 +224,13 @@ examples:
         {{ summary.field_name("Support available") }}
         {{ summary.boolean(True, "CESG-assured components") }}
       {% endcall %}
+      {% call summary.row() %}
+        {{ summary.field_name("Developer") }}
+        {{ summary.multiquestion([
+          {'type': "list", 'value': ["London", "Mississauga", "Rivière-du-Loup", "Singapore"]},
```
TheDoubleK: :maple_leaf:

TheDoubleK:
All looks nice now, :+1: from me, although maybe @quis could take another quick look before merging.

screen shot 2015-11-23 at 11 50 57

The templating is looking pretty involved these days...
screen shot 2015-11-23 at 11 52 07

quis:
Checked out locally and nothing is broken any more so :+1: from me.

PR #185, opened 2015-11-19
Multi message spacing

TheDoubleK:
Not actually an issue right now, but could be at some point. When there are multiple messages on the supplier dashboard they look too bunched up (see screenshots below).

I have no idea if this is the best approach to fix it, but it seems to make sense to add a margin at the bottom of one element type rather than a top margin to all the others.

Before:
screen shot 2015-11-19 at 13 48 24

screen shot 2015-11-19 at 13 48 39

After:
screen shot 2015-11-19 at 14 12 56

screen shot 2015-11-19 at 14 12 37


toolkit/scss/_temporary-message.scss
```diff
@@ -9,6 +9,7 @@
   background: $govuk-blue;
   color: $white;
   padding: 15px 15px 18px 15px;
+  margin-bottom: 20px;
```
quis: Can this be `$gutter` (30px) instead? Sorry for pedantry.

TheDoubleK: Yep no worries.  What's the difference between a gutter and a margin?

pcraig3: I think he means do `margin-bottom: $gutter;` where `$gutter == 30px`.  

Not sure when we use paddings and when we use margins though, if that's your question.

quis:
Paul is right. 

And technically, margins are the space around the edges of a page, gutters are the space between columns of text. But CSS calls them all the wrong things so `¯\_(ツ)_/¯`

TheDoubleK:
Fixed up.

quis:
:+1:

PR #187, opened 2015-11-24
Change `text` macro call to `text_content`

pcraig3:
Our text macro is in a different file, so it wasn't defined.
Use `text_content` instead.  Easy.


allait:
:+1:

PR #188, opened 2015-11-24
Make pricing adaptable

robyoung:
## Allow pricing fields to be set in the data
The fields the pricing module is made of can now be set in the template data.

The new keys are non longer in camelCase to better match the style the rest of our python code has.

## Fix pricing spacing issue when validation fires
The pricing module has a negative margin-right on desktop to allow it to stretch out of the two-thirds column.

When the validation fired, it wrapped the module in a `.validation-wrapper` div which constrained the width to the parent column.

This adds the same negative margin to the `.validation-wrapper` div to allow its contents to stretch.

## Allow price fields to be marked as optional


quis:
Checked out locally and :+1:

PR #189, opened 2015-11-26
Make browse list status lists inherit colour

quis:
| Before | After |
| ----------| ---------- |
| image | image |

This is so that:
- the visual grouping is stronger
- the lists aren’t easily confused with the hints
- it matches [what the prototype does with ‘You’ve completed…’](http://dm-prototype.herokuapp.com/states/framework-dos?dos.framework=open&dos.application=null)

Also tweaks the spacing to make it look more balanced


ralph-hawkins:
Can’t speak for the code, but :+1: +1 with the design change.

PR #190, opened 2015-11-27
Selection buttons now contain their description

ralph-hawkins:
If a selection button has a description with it, the description now sits inside the selection button.

**Old:**
screen shot 2015-11-27 at 10 10 14

**New:**

screen shot 2015-11-27 at 10 09 35

This is better because the description is more strongly associated with the thing being selected and when there are a lot of options with descriptions the page is doesn’t zebra stripe down the page.


quis:
Seems legit. You should also [bump the version number to 13.0.2](https://github.com/alphagov/digitalmarketplace-frontend-toolkit#versioning).

PR #191, opened 2015-11-27
Thicker textbox borders

quis:
# Before

image

# After

image

--

image

--

image

## Make borders around textboxes thicker

The original, 1px grey borders have caused problems in (other services’) testing with users who have poor eyesight.

This commit follows the updated http://govuk-elements.herokuapp.com/form-elements/ which has a thicker, darker border.

## Make inset text match border colour

This changes the colour of labels that live inside textboxes, eg numbering them, adding a unit.

Before this change there were too many shades of grey.


allait:
:+1:

PR #193, opened 2015-11-30
Fix broken links on homepage

quis:
When the nested links were added to the homepage in https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/181 their (relative) paths were not updated accordingly.

This commit fixes the paths so the links are not broken.


TheDoubleK:
:+1:

PR #194, opened 2015-11-30
Add optional ID to summary heading pattern

quis:
Pages with summary tables can be quite long.

You may wish to link to an anchor some way down the page.

Adding the option to pass an ID to the table heading makes this possible.


allait:
:+1:

PR #195, opened 2015-11-30
Re-arrange hours and pricing

carolinegreen:
Re-arrange hours and pricing as it makes more sense to ask the minimum amount of time that can be booked before the price is asked about.


allait:
:+1:

PR #197, opened 2015-12-01
Stop question numbers from overlapping questions

quis:
Most of the way towards fixing https://www.pivotaltracker.com/n/projects/997836/stories/106191644

--

Desktop
image

--

Mobile
image

--

Desktop
image

--

Mobile
image

--

On < desktop sized screens the question numbers in the supplier declaration overlap the text of the questions.

This commit changes the question numbers so that:
- they are only outdented on the very largest screen size
- they are otherwise displayed inline with the question

This also adds question numbers to some of the example forms so that the pattern can be tested more easily.

This showed up a problem when question numbers were wrapped in a validation error, which is also addressed by this commit.


ralph-hawkins:
:+1:

PR #198, opened 2015-12-01
back link pattern

ralph-hawkins:
For when you need to get back to the previous page.


quis:
giphy

quis:
You should squash the last two commits together.

quis:
:+1:

PR #200, opened 2015-12-03
Make question headings regular weight

quis:
Bold question headings look shouty when the questions are long (eg service
level) declaration questions.

This commit changes all question headings to be light, but the same size. This
goes against what’s in GOV.UK elements, but works for our pages with multiple,
complex questions, as well as our unique patterns (like pricing).

--

image

image

image

image

image


allait:
:+1:

PR #202, opened 2015-12-08
Improve support for empty fields in summary table cells

tombye:
This improves the support for empty cells in summary tables in the following ways:

1. ensures the macros that add formatting don't do it to empty strings
2. doesn't wrap empty values in a `<span>` inside the cell

Number 1. should allow the fixing of bugs like this one on the supplier app: https://www.pivotaltracker.com/story/show/109618680.

#### Notes

I've added an example to the summary-tables page.

All call blocks sent to the `field` macro have stricter whitespace control added. This ensures the HTML string sent in comes through as empty if that's what the caller macro returns.


quis:
I think it’s a bit weird to suggest that it’s OK to call these macros with missing arguments. But since they’re documented with the examples it’s probably OK. 

:+1:

toolkit/templates/summary-table.html
```diff
@@ -83,83 +83,101 @@ <h2 class="summary-item-heading"{% if id %} id="{{ id }}"{% endif %}>
...
   </td>
 {%- endmacro %}
 
-
 {# SPECIAL FIELD TYPES #}
 {# =================== #}
 
-{% macro service_link(service_name, link, wide=True) -%}
+{% macro service_link(service_name=None, link=None, wide=True) -%}
   {% call field(first=True, wide=wide) %}
```
quis: Missing the whitespace control here.

PR #203, opened 2015-12-14
Bump version to 13.4.3

tombye:
`13.4.2` was inserted in the middle of https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/202 so this brings in those changes properly.


tombye:
@quis just :+1: this in 'real-life'.

PR #205, opened 2015-12-17
Bump option-select to latest

tombye:
Brings in the following changes:

- [Accessibility fixes](https://github.com/alphagov/static/pull/592)
  - make the bit you click use a `<button>` tag (so assistive tech users are told they can click it) and adds semantics describing the relationship between the button and the options container below (using [WAI-ARIA](http://www.w3.org/WAI/intro/aria))
- [Add the ability to set option-selects as closed when the page loads](https://github.com/alphagov/static/pull/594)

(Includes fixes for the problem described in this story: https://www.pivotaltracker.com/story/show/105291876


TheDoubleK:
Looks good :+1:

tombye:
Missed a few changes that were rolled into the updates on the govuk component. Now included.

PR #206, opened 2015-12-18
Stop document links with icons from overflowing

pcraig3:
Document links with icons have padding applied to them which pushes them out of their parent containers. This led to a bug on smaller screens where the page layout would be pushed beyond the edge of the screen.
Setting `box-sizing: border-box;` on the link elements resolves this.

***

## Bad
screen shot 2015-12-18 at 17 16 47

## Good
screen shot 2015-12-18 at 17 15 30


tombye:
:+1:

PR #210, opened 2016-01-04
Add history status banner

tombye:
As part of the story to keep old services on the marketplace, this is the pattern used on www.gov.uk for marking a page as referring to a resource bound to a period of time (and other factors).

image

(See https://www.pivotaltracker.com/story/show/110062348 for details of the story and this [report on the 2012 olympics](https://www.gov.uk/government/news/turning-the-games-into-gold-economic-boost-from-london-2012) for an example of the pattern on www.gov.uk.)


quis:
I can’t see the difference between this and http://alphagov.github.io/digitalmarketplace-frontend-toolkit/temporary-message.html





<sup>sorry :crying_cat_face: </sup>

tombye:
@quis quite right. I've added it as a temporary message but this needed a few bigger changes:

1. the addition of a `temporary-message-full-width` class to allow larger text sizes
2. some changes to how grids are used in the example pages to allow a full-width example

For 2. see e8c26ab and f6aa262 for details.

I also added some ARIA to allow us to associate the message heading with it's container `<aside>`.

tombye:
OK, so it turns out this should actually be a style of notification banner.

I've added an example to that page and split the `%temporary-message` placeholder off into a separate file so it can be shared between this new style and those in the `toolkit/scss/_temporary-messages.scss` file.

pages_builder/pages/notification-banner.yml
```diff
@@ -27,3 +27,12 @@ examples:
     title: Destructive banner
     type: destructive
     message: <a href="#">With a link</a> and some plain text but without an action
+  -
+    title: Temporary message banner
+    type: temporary-message
+    heading: >
+            <h2>
```
quis: We should avoid passing HTML to patterns wherever possible (and I can’t see what the `history-status-date` class is being used for.

tombye:
There's too much in this pull request for easy reading (history and features).

I'm closing it in favour of https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/211 for the feature and https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/212 for the changes to temporary messages.

PR #212, opened 2016-01-06
Fixes to temporary messages

tombye:
Add some ARIA to label the `<aside>` sections (using [this technique from the styleguide](https://github.com/alphagov/styleguides/blob/master/html.md#sectioned-content)).

This includes some changes to the toolkit to allow pattern instances to each exist in their own grid column.

### DIFF
--- a/pages_builder/generate_pages.py
+++ b/pages_builder/generate_pages.py
@@ -167,11 +167,18 @@ class Styleguide_publisher(object):
                         example_template = example
                         example_markup = env.from_string(example_template).render({})
 
+                    # set a grid if specified. Example-level grids will overwrite the one for the page
+                    grid = None
+                    if 'grid' in example:
+                        grid = example['grid']
+                    elif 'grid' in partial:
+                        grid = partial['grid']
+              
                     examples.append({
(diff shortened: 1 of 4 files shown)


pages_builder/generate_pages.py
```diff
@@ -167,11 +167,18 @@ def render_page(self, root, file):
                         example_template = example
                         example_markup = env.from_string(example_template).render({})
 
+                    # set a grid if specified. Example-level grids will overwrite the one for the page
+                    grid = None
+                    if 'grid' in example:
+                        grid = example['grid']
+                    elif 'grid' in partial:
+                        grid = partial['grid']
+
```
quis: More pythonic (:trollface:):
``` python
grid = example.get('grid') or partial.get('grid') or None
```

robyoung: (even) more pythonic
```python
grid = example.get('grid', partial.get('grid'))
```

tombye: Nice!!! (now updated.)

robyoung:
:+1:

PR #211, opened 2016-01-06
Add service unavailable banner

tombye:
Add a notification banner for services no longer available to buy.

https://www.pivotaltracker.com/story/show/110062348

image

Note: this is a re-vamped version of https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/210

PR #213, opened 2016-01-11
Bump version to 15.0.0

tombye:
Includes

- new notification banner format
  [#211](https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/211)
(breaking change)
- changes to temporary messages example page
  [#212](https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/212)
- separation of temporary message style into
  placeholder [in
- ability to add pattern-level grids [in
  #212](https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/212)

### Notes

The new notification banner commit introduced a breaking change. 

The base temporary message styles were split out into a placeholder block to allow use across multiple patterns (by [extending](https://github.com/alphagov/digitalmarketplace-frontend-toolkit/blob/master/toolkit/scss/_notification-banners.scss#L69) it):

https://github.com/alphagov/digitalmarketplace-frontend-toolkit/blob/master/toolkit/scss/shared_placeholders/_temporary-messages.scss

This means the existing temporary message pattern will break unless the new placeholder block is included in the main stylesheet of apps that use it.


TheDoubleK:
Do we want to add a changelog to the toolkit, as we have been introducing elsewhere?
That notwithstanding :+1:  for the version bump.

tombye:
@TheDoubleK Yep, that's a good idea. Now added.

PR #214, opened 2016-01-15
Fix issue with new grids setting

tombye:
Some pages have example data which is a string while most are dictionaries. The code that checked
for a grid being set per-example assumed they were dictionaries so was breaking.


robyoung:
:+1:

PR #216, opened 2016-01-18
Override default link colours in temporary messages

pcraig3:
The default link colour is the same colour as the background colour of a
temporary message, effectively rendering links invisible until hovered over.
New CSS colours links white in temporary messages, removing the underline
when hovered over.
Approved by @ralph-hawkins.

***

## Before
screen shot 2016-01-18 at 18 45 28

## Before:hover
screen shot 2016-01-18 at 18 42 52

## Now
screen shot 2016-01-18 at 18 42 15


VERSION.txt
```diff
@@ -1 +1 @@
-15.0.1
+15.1.0
```
TheDoubleK: I might argue that this is a bugfix and so the new version should be 15.0.2.  But it really doesn't matter.

toolkit/scss/shared_placeholders/_temporary-messages.scss
```diff
@@ -6,4 +6,13 @@
...
   color: $white;
   padding: $gutter-half $gutter-half 18px $gutter-half;
+
+  a, a:visited {
+    color: $white;
+  }
+
+  a:hover, a:active {
+    @extend a;
+    text-decoration: none;
```
TheDoubleK: Does this remove the underline when the link is hovered over?  I don't think that happens with any of our other links - they change colour but remain underlined.  It seems odd to change that pattern just for these links?

ralph-hawkins: The links usually go from `$link-colour` to `$link-hover-colour`, which we can’t use because of the similarity in background colour. To show on hover that the underlined text is a link we could either change from `white` to `$grey-4`or `text-decoration: none;` the later was my preference, but maybe colour change would be more consistent.

TheDoubleK: It's your call - underline going might look nicer, but not sure if consistency across the site trumps that?

quis: links

ralph-hawkins: agreed – go to shade or two darker on hover

pcraig3:
`$grey-4` is too light, so `$grey-3` seems like the better option.
gov.uk are going with `#ccc`, but that's not part of their [`_colours.scss`](https://github.com/alphagov/govuk_frontend_toolkit/blob/master/stylesheets/_colours.scss) file, so I don't know what's going on there.

## Now:hover
screen shot 2016-01-19 at 11 49 19

VERSION.txt
pcraig3: Seems like a reasonable suggestion.

quis:
`$light-blue-25`

https://en.wikipedia.org/wiki/Tints_and_shades

#knowledgetransfer

pcraig3:
Not sure I like your tone, @quis.

quis:
Sorry for acting shady.

pcraig3:
http://www.urbandictionary.com/define.php?term=throwing+shade

> Throwing Shade, is to throw "attitude."

TheDoubleK:
:+1: from me.  But what do I know?

ralph-hawkins:
:+1:

PR #218, opened 2016-01-28
Add button attribute

carolinegreen:
Add button attribute to button element to fix searchbox bug in buyer frontend where pressing
enter on textbox trigger synthetic mouselick event on nearest button in the form.  Adding type button
to the button element overrides default the button default type of submit within the form.


tombye:
Great sleuthing :+1:

PR #220, opened 2016-02-08
Fix Makefile when user is not in a virtualenv

robyoung:
We need to use absolute paths to executables that need to be used from
the virtual environment to work for the case when the user is not
already in a virtual environment.


allait:
:+1:

PR #221, opened 2016-02-29
Smaller textboxes

TheDoubleK:
For the respond-to-a-brief questions "start date" should be a small text box, which until now was impossible without specifying a unit.


pcraig3:
<sub>+1</sub>

PR #222, opened 2016-03-02
Change link button margin

tombye:
We're now only using this at the foot of long-form pages like the 'view-brief' page so the margin-left isn't needed.

Before:

link-button-old

After:

link-button-new


allait:
:+1:, do we need a version bump?

tombye:
Yep, gonna do a PR for the version bump now.

PR #223, opened 2016-03-03
Re-add small textboxes that were lost

TheDoubleK:
Re-introduce small textboxes from v15.2.0, which we still need.


TheDoubleK:
Tom is going to work some Github sorcery to bring these back the right way.

PR #225, opened 2016-03-03
Add pattern for instructions list

carolinegreen:
Add pattern for instructions list.  I am keeping this separate from the browse list pattern, as the browse list is navigational and would group better with other navigational patterns.

Currently fields within the instruction list which borrow from other patterns such as uploads and documents are being included using the with extension.  This extension has also been enabled for the frontend toolkit itself.  Ideally these patterns would be included as macros to allow greater flexibility.  However, as this would be a breaking change it is probably better to change these patterns into macros gradually and in a separate commit.

A download option has been added to the document pattern, to allow documents to be downloaded.

Lists have a "start" attribute included to allow them to be broken into sections and other components to be introduced between blocks of lists if required.  Lists also have a 'verbose' option for lists which contain sublists or a large amount of text and require more obvious headers and dividers to break up each item.

Ideally the lists CSS could be removed from the supplier app entirely, rather than being periodically removed and then re-introduced, and stored in the frontend toolkit instead.

Default List:

screen shot 2016-03-03 at 23 32 55

Verbose List:

screen shot 2016-03-03 at 23 33 57


pages_builder/pages/index.yml
```diff
@@ -47,6 +47,7 @@ content: >
           </ul>
         </li>
         <li><a href="framework-notice.html">Framework notice</a></li>
+        <li><a href="instruction-list.html">Instruction list</a></li>
```
TheDoubleK: `I` comes after `G` in the alphabet ;)

toolkit/scss/_instruction-list.scss
```diff
@@ -0,0 +1,105 @@
+@import "_colours.scss";
+@import "_typography.scss";
+@import "_measurements.scss";
+@import "_shims.scss";
+@import "_conditionals.scss";
+
+ul,
```
TheDoubleK: Are these intended for all `<ol>` and `<ul>` in the application, or should they be wrapped in a class name or something?  It feels like styles here should be applied specifically to toolkit elements only, not as a blanket for the whole app.
We have in-app `reset` styles like this: https://github.com/alphagov/digitalmarketplace-buyer-frontend/blob/master/app/assets/scss/_reset.scss

toolkit/scss/_instruction-list.scss
```diff
@@ -0,0 +1,105 @@
...
+ol {
+  list-style-type: none;
+  padding: 0;
+}
+
+// Bulleted lists
```
TheDoubleK: We also have these styles pulled in from gov.uk elements, e.g.: https://github.com/alphagov/digitalmarketplace-buyer-frontend/blob/master/app/assets/scss/_lists.scss

toolkit/scss/_instruction-list.scss
```diff
@@ -0,0 +1,105 @@
...
+      background-size: 34px 34px;
+    }
+  }
+
+  strong {
+    display: block;
+    padding-left: (35 + 30)+px;
+    margin-left: -$gutter-half;
+  }
+}
```
TheDoubleK: I know Javascript doesn't mind too much about end of file line breaks, but they don't hurt.

TheDoubleK:
When I do a `make serve_pages` locally the warning image is cropped - not sure if this is something I've done wrong locally though?
screen shot 2016-03-04 at 10 42 21

carolinegreen:
@TheDoubleK No it's a bug.  I pulled this in from elements.  However didn't notice it as I have a retina screen.  It's good you spotted it, I will open a pull request for that as well.

pages_builder/pages/index.yml
carolinegreen: Yes, that is true :o

toolkit/scss/_instruction-list.scss
carolinegreen: Yeah

toolkit/scss/_instruction-list.scss
carolinegreen: I took this from the list scss in the supplier app, but yes you're right, it is, and probably has been redundant for a while.

toolkit/scss/_instruction-list.scss
carolinegreen: That's the plan :)

TheDoubleK:
Nice - all looks good now :+1:

PR #229, opened 2016-03-11
Remove padding from last upload list item

carolinegreen:
Remove padding from last upload list item and use css selectors instead classes for dividers. Also fix inconsistencies in naming.


tombye:
:+1:

PR #230, opened 2016-03-17
Isolate some shared styles not connected to macros.

pcraig3:
For some of our error pages, we want to have access to the same CSS classes but we don't necessariy need corresponding macros.

This pull request isolates some of our existing list styles (and adds more), as well as adds a `.dmspeak` class that, at this point, adds more spacing between paragraphs.

Once this is merged, we can remove `lists.scss` from the buyer app (it's already been removed from the supplier app).


carolinegreen:
I'm still not convinced that explanation list and check list make sense without examples and some kind of context as they have enough rules that it looks like some kind of structure is expected.  Coming to it without seeing a visual representation my immediate question would be an explanation of what?  

Putting the other styles into a shared section makes sense though.

allait:
I think we can add an pure HTML example to the documentation without creating a macro/Jinja include template.

The overall structure seems simple enough to repeat and explanation list contents would probably contain other HTML elements like links most of the time, which is a big issue with our current patterns right now (combining HTML into strings so that it can be passed in, having to mark it as safe – which means we have to be careful not to use any user input in the contents etc.) so I don't think a pattern include would make things better in this case.

carolinegreen:
If we're going to do that then it would probable be better for an html example be included with this pull request, or a template and a note that we will be replacing the simpler patterns with html examples, rather than leaving it inconsistent with classes that don't really have a context.

pcraig3:
Added a page with examples of usage.

***

#### http://localhost:8000/shared_scss/_lists.html
screen shot 2016-03-18 at 15 46 16

pages_builder/pages/shared_scss/_lists.yml
```diff
@@ -0,0 +1,88 @@
...
+        1–1½ oz. vodka
+      </li>
+      <li>
+        6 oz. clamato juice
+      </li>
+    </ul>
+  - |
+    <h2><code>.list-number</code></h2>
+    <br />
+    <ul class="list-number">
```
allait: This probably should be `<ol>`

pcraig3: <ol>
 <li>oops.</li>
 <li>i'll fix it</li>
</ol>

allait:
:+1:

tombye:
👏

pcraig3:
:metal:

ralph-hawkins:
Is there additional spacing on the last list item on each example?

tombye:
@ralph-hawkins the numbered and bulleted lists don't but the checklist does:

https://github.com/alphagov/digitalmarketplace-frontend-toolkit/blob/master/toolkit/scss/shared_scss/_lists.scss#L51

Should there be?

ralph-hawkins:
It looked like the last item on each list was a bit further away than the others, but I think its just my eyes, ignore me.

PR #231, opened 2016-03-17
Fix instruction list parargraph margins

carolinegreen:
Add margins to instruction list paragraphs as these are too close together in the short version if a css reset is present.


tombye:
:+1:

PR #232, opened 2016-03-18
Fix `list` listing in index list.

pcraig3:
[Jeykl ignores files that start with underscores](https://help.github.com/articles/files-that-start-with-an-underscore-are-missing/), apparently.  [Like this one](http://alphagov.github.io/digitalmarketplace-frontend-toolkit/shared_scss/_lists.html).

After close examination, it seems like just because I added a `_lists.scss` file, I didn't also need a `_lists.yml` file.  In fact, all of the other `.scss` files start with underscores, but none of the other `.yml` files start with underscores. 

<sub>Whoops.</sub>


allait:
:+1:

PR #234, opened 2016-04-12
Add list-bullet style copy to the question hint <ul>

allait:
Since question hints are processed with markdown they can contain lists that should be displayed with bullet points, similar to how they're shown in the prototype.

screen shot 2016-04-12 at 17 38 23


TheDoubleK:
It might be nicer to re-use the toolkit `list-bullet` style itself as an extend/include rather than duplicating the individual css.

allait:
@TheDoubleK updated with a mixin include to reuse the CSS block

pcraig3:
:+1: 

thumbs_up

PR #235, opened 2016-04-13
Split up field hints

tombye:
We have a need for the hints that appear below field labels to be split into two different types:

1. advice giving more information about the question
2. advice on how to answer the question

image

I've added it in a non-destructive way by retaining the `hint` property you send into the jinja templates but adding the new hint type as `question_advice`. At some point in the future we should change all the `hint` properties to `answer_advice` but this will do for now.

This includes quite a few changes because I've added some semantics (using `aria-describedby`) to identify the two advice types as help text for the field control. Ideally, the advice text for questions should be linked to the field label but after testing with the OSX accessibility api, browsers didn't create the association whereas they did when it was linked to the field control (where it is added as an `accessibilityHelp` attribute).


carolinegreen:
Yes, I think it's good to go :+1:

PR #236, opened 2016-04-18
Add new options for summary table

carolinegreen:
Add new options for summary table based on design pattern in prototype kit:
  - New narrow class, used for dates in example.
  - Different border options.
    -  No-top-border option, bold border for rows.
    -  Light border for table data.
    -  No bottom border option for entire table.
  - Large text - for text with a font size of 19 in summary tables.
  - Lead in pararaph - for paragraph above table that isn't a description.


toolkit/scss/forms/_summary.scss
```diff
@@ -46,7 +49,7 @@
 
 .summary-item-no-content {
   color: $secondary-text-colour;
-  @include core-16;
+  @include copy-16;
```
tombye: `copy-16` brings in `padding-top` and `padding-bottom` so do we still need those 4 lines down from here?

toolkit/scss/forms/_summary.scss
```diff
@@ -179,7 +260,6 @@ a.summary-item-remove-link {
 
     span {
       display: block;
-      padding-left: 10px;
```
tombye: Taking this `padding-left` out seems to have broken something in the example page:

image

toolkit/scss/forms/_summary.scss
carolinegreen: It looks pretty horrible with the different padding.  I could just go back to core-16, as I'm not sure the fallback for tablets would look great anyway with the padding not being great on full screen view.

toolkit/scss/forms/_summary.scss
carolinegreen: Yeah, got too enthusiastic about removing padding from everything.  Adding it back in.

tombye:
Thanks for changing the things mentioned.

I think there are a few things I've noticed that still need changing:

1. the first cell in each group of rows (most of which contain the date for the group) should be a `th` rather than a `td` so it's identfied as the heading for them
2. after talking to @ralph-hawkins about the borders on another summary table, it looks like we should remove the `border-top: solid 1px black` style added here for now

### Point 1

This will probably need the creation of a new macro for this (perhaps `heading_field_date`?) which lets us set the `scope` and `rowspan` properties. Doing so should mean it can be associated with the rows it needs to be (as in this example: http://htmldog.com/examples/rowspan.html)

### Point 2

Ralph is going to introduce a DM toolkit-level `$border-colour` variable and when that happens we can use it here for the darker borders but until then he's ok with their colour being the same as the others (with class hooks for later work).

carolinegreen:
1. I've removed the headers and ids properties and replaced them with scope, and replaced td with th as suggested.
2. I've replaced black with the $black mixin as I think there still needs to be a contrast between the border colours.

tombye:
👍

PR #237, opened 2016-04-19
Add additional markdown filter for framework notice subheadings.

pcraig3:
Added a `|markdown` filter for framework notice subheadings, updated version of dmutils to a new one which supports [markdown abbreviations](http://pythonhosted.org/Markdown/extensions/abbreviations.html), as well as I updated the framework notice subheading to something with a markdown abbreviation.

Nothing from dmutils appears to be being used elsewhere, so no worries about the big jump in versions.

Finally, I changed `>` to `|` and remove unneeded linebreaks in the framework notice example file.


allait:
👍

PR #239, opened 2016-04-20
Add Optional block, style advice blocks

pcraig3:
#### 1. (optional) Optional blocks
Adds an (optional) **Optional** block to questions similar to [what we see in the prototype](http://dm-prototype.herokuapp.com/forms/checklist?config=session-times&nextPage=/participants-brief). 
This means we can stop using hint text to indicate questions are optional

#### 2. styles for `.question-advice` and `.hint` classes
Adds a bunch of CSS to `.question-advice` and `.hint` classes, almost entirely cribbed from the buyer app's [`.section-description`](https://github.com/alphagov/digitalmarketplace-buyer-frontend/blob/master/app/assets/scss/_brief_submission.scss#L55) class which is used on section summary pages. 
Once this gets merged, we can remove the duplicated stuff from the buyer app.

#### 1. fixing the duplicated questions
At the moment, the question itself appears duplicated on the edit question page ([here's a screenshot](https://cloud.githubusercontent.com/assets/2454380/14681749/6095f3d0-071a-11e6-891e-7068f2d7602f.png)). We need to pass the question into the macros because they become the `<label>` for the form fields, but the question is also used as a page heading for individual question pages.  Currently, we need to do both.

#### 2. moving hint text
In the prototype, [hint text has been moved underneath form fields](http://dm-prototype.herokuapp.com/forms/textbox?config=brief-name&nextPage=/outcome-overview).  I haven't touched it.

#### 3. lead-in values for list questions
In the prototype, list questions have a short phrase as a heading ([ie, "The supplier must:"](http://dm-prototype.herokuapp.com/forms/list-entry?config=compulsory-qs&nextPage=/outcomes-evalutaion)).  Doesn't look immensely difficult to add in, but I didn't do it.

***

#### optional question with tons of question advice, a hint, and an error

screen shot 2016-04-20 at 16 56 41


allait:
LGTM 👍

PR #240, opened 2016-04-21
[hotfix]  add back `%summary-item-row` partial 

pcraig3:
Should fix this problem: 
screen shot 2016-04-21 at 11 21 00


allait:
👍

PR #241, opened 2016-04-21
Keep summary table text at 16px

pcraig3:
Summary table text was bumped up to 19px everywhere, but it doesn't really look that awesome.  (slash we haven't really looked at everywhere it's used to make sure that it works.)

Keeping it at 16px for time being.

***

### 16px (what we are used to)
screen shot 2016-04-21 at 12 34 07

### 19px
screen shot 2016-04-21 at 12 33 58


ralph-hawkins:
At some point soon we make published opportunities at 19px.

carolinegreen:
Actually I think you can remove the entire thing if you remove all the uses of this in other sections, which was what I didn't do.

carolinegreen:
👍

PR #242, opened 2016-04-21
Fix framework notice message

pcraig3:
Fixes two problems with framework notices:

- list bullets were unintentionally removed during the [Great List Rationalization](https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/230)
  - fixed by adding them back in manually
- markdown filter creates a new `<p>` tag without a classname.
  - fixed by wrapping it in a `<span>`

***

### [current](http://alphagov.github.io/digitalmarketplace-frontend-toolkit/framework-notice.html) 😢 
screen shot 2016-04-21 at 14 54 07

### new 😃 
screen shot 2016-04-21 at 14 54 41


TheDoubleK:
Looks moar better :+1:

PR #243, opened 2016-04-21
added padding-right: 12px; to .summary-item-field-first

ralph-hawkins:
so it doesn't crash into the next column if content is the same width as the column


allait:
👍

PR #245, opened 2016-04-22
Fix validation copy

tombye:
As requested by @superroz.


TheDoubleK:
There is a possibility this will break the functional tests, although they are mainly happy-path so maybe not. Worth a check before we pull this version into ant apps though.

:+1: for the change here.

tombye:
Good call, I'll be sure to run them first.

tombye:
I've run them locally and they seem happy (I think that's the problem, it's all happy-path in there).

PR #247, opened 2016-04-22
Minor refinements to question templates

pcraig3:
## 1. Remove bottom padding from question_advice and hint fields

A padding-bottom of 10px was being added to markdown formatted
text which we don't really need because there's a padding-bottom
of 10px added to all paragraphs.

Removing the padding-bottom makes questions with both question_advice
and a hint look neater.

#### --> before 
screen shot 2016-04-22 at 17 57 03

#### --> after
screen shot 2016-04-22 at 17 56 13

***

## 2. Preserve margins on validation messages

We were losing vertical margins on `.validation-message`s because the parent container wasn't expanding to account for them. [Learn more about block formatting context (BFC).](http://colinaarts.com/articles/the-magic-of-overflow-hidden/)  <sub>disclaimer: not that interesting</sub>

#### --> before
screen shot 2016-04-22 at 17 47 59

#### --> after
screen shot 2016-04-22 at 17 48 30

### DIFF
--- a/toolkit/scss/forms/_validation.scss
+++ b/toolkit/scss/forms/_validation.scss
@@ -53,7 +53,7 @@
 }
 
 .validation-message {
-  display: block;
+  @include inline-block();
   @include core-19;
   font-weight: bold;
   color: $red;
--- a/toolkit/scss/shared_scss/_dmspeak.scss
+++ b/toolkit/scss/shared_scss/_dmspeak.scss
@@ -10,8 +10,7 @@
 }
 
 %markdown-description {
-   @include core-19;
-   padding-bottom: 10px;
+  @include core-19;
 
   p {
     margin: 0;


toolkit/scss/forms/_validation.scss
```diff
@@ -53,7 +53,7 @@
 }
 
 .validation-message {
-  display: block;
+  display: inline-block;
```
tombye: It'd be worth using the [inline-block shim](https://github.com/alphagov/govuk_frontend_toolkit/blob/master/stylesheets/_shims.scss#L17) from the govuk toolkit here.

tombye:
I'm not clear we do add 10px of padding-bottom to paragraphs. I can't find that in any of the CSS resets (for the [buyers](https://github.com/alphagov/digitalmarketplace-buyer-frontend/blob/master/app/assets/scss/_reset.scss), [suppliers](https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/master/app/assets/scss/_reset.scss) or [admin](https://github.com/alphagov/digitalmarketplace-admin-frontend/blob/master/app/assets/scss/_reset.scss) apps). There is a 1em margin in the default browser styles but we shouldn't rely on it.

pcraig3:
`padding-bottom: 10px;` is applied to paragraphs in `%markdown-description` blocks (ie, question hints and question advice).  [See here.](https://github.com/alphagov/digitalmarketplace-frontend-toolkit/blob/master/toolkit/scss/shared_scss/_dmspeak.scss#L18)

ralph-hawkins:
👍

PR #248, opened 2016-04-26
Add temporary message variant for main pages

carolinegreen:
Add temporary message variant for main pages.

screen shot 2016-04-26 at 17 50 29


ralph-hawkins:
:+1:

ralph-hawkins:
Although it'd be preferable to use  `G&#8209;Cloud&nbsp;8` instead of 'G-Cloud 8', however, in this case it's unlikely to matter.

TheDoubleK:
No need to go all-out with the non-breaking just for the toolkit example. I've put a thing in Pivotal to look at sorting it for the actual app content.

ralph-hawkins:
You're right! Sorry, forgot this was toolkit.

PR #250, opened 2016-04-28
Bump version to 15.12.1

pcraig3:
I had thought this would just be pulled in with another change, but it doesn't look like anything else is going through right now and it's worth pulling these small layout fixes into the buyer app explicitly.

Brings in #247.


TheDoubleK:
:+1:

PR #251, opened 2016-04-28
Rewrite temporary message inheritance

carolinegreen:
Remove temporary message section, and assign this to the temporary-message class which is clearer.  Also add inheritance from this variable to temporary-message-main class so that class has correct margin.


toolkit/scss/_temporary-message.scss
```diff
@@ -7,18 +7,14 @@
...
 
-%temporary-message-section {
-  @extend %temporary-message;
-  margin-bottom: $gutter;
-}
-
+%temporary-message-section,
 .temporary-message {
-  @extend %temporary-message-section;
+  @extend %temporary-message;
```
pcraig3: Do you think it's worth importing this in this file?  

As it stands, it seems like a hidden dependency on the `shared_placeholders/_temporary-messages.scss` file.

carolinegreen: Yes, it this import wasn't something I added, and other files are using this.

pcraig3: I mean, do you think it's worth adding an `@import "shared_placeholders/_temporary-messages.scss";` statement at the top of this file?

carolinegreen: I think the problem is that extend is being used and if the shared_placeholders.scss file is imported again further down by another file, the changes in the extended class will be overwritten.  This has already surfaced as a problem in the buyer app, where the shared sass placeholder partial was introduced after the temporary message partial.

I think the best thing is probably to leave it just now, but I will put a story in pivotal tracker for looking at how imports are being handled in general.  Even imports are left as they are, it probably is still worth putting in documentation to explain the decisions behind how they are being handled.

pcraig3:
👍

toolkit/scss/_temporary-message.scss
pcraig3: Ah, makes sense.  Sounds a bit convoluted, but w/e.

carolinegreen: Imports still aren't great in sass sadly.

carolinegreen: The way sass handles imports still isn't great unfortunately.

PR #253, opened 2016-05-03
Bring in _forms.scss styling to toolkit

pcraig3:
Two of our three apps have duplicate `_forms.scss` files ([the buyer app](https://github.com/alphagov/digitalmarketplace-buyer-frontend/blob/master/app/assets/scss/_forms.scss), and [the supplier app](https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/master/app/assets/scss/_forms.scss)), and instead of adding it to the admin app (to correct the correct spacing above the log in button), it made more sense to bring the styles into the toolkit.

Once this is merged, we can delete the duplicated CSS in the other two apps.


allait:
👍

PR #254, opened 2016-05-03
Keep validation messages on their own line

pcraig3:
Fixes a bug where validation messages would share a line with input elements if those input elements weren't block-level elements.

Reverting validation messages back to block-level elements and changing the `margin`
to `padding` will:

- preserve spacing
- keep input fields on their own lines

***

### yuck
screen shot 2016-05-03 at 13 31 24

### good
screen shot 2016-05-03 at 13 32 03


ralph-hawkins:
👍 👍 👍

PR #255, opened 2016-05-04
Rename percentage question summary templates to 'number'

allait:
Renames summary.percentage to summary.number and stops the template from appending '%' sign to the value. The correct unit should be added by `ContentQuestionSummary.value`.


TheDoubleK:
:+1:

PR #256, opened 2016-05-09
Top margin only applied to <input> elements (not <button> elements)

pcraig3:
Bugfix for the alignment bug on the search button on G-Cloud.

Only targets `<input>` elements with the `button-save` class.
This the original style in [`_forms.scss`](https://github.com/alphagov/digitalmarketplace-buyer-frontend/blob/67b6834c7bab9d1a18b3b90b5254f772c11b5169/app/assets/scss/_forms.scss#L2) in the buyer app, ([relevant pull request](https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/253)) but I generalised it because it seemed like it was fine.
Obviously not.


carolinegreen:
Looks fine 👍

PR #258, opened 2016-05-10
IE8 doesn't support `bind` natively so shim it

tombye:
The JS for selection-buttons uses `Object.prototype.bind` which is supported in Internet Explorer from version 9 upwards. Because of this, whenever we use it, we need to include the shim for `bind` which adds support for <=IE8.

It's annoying debugging toolkit patterns in IE at the moment because you get JS errors thanks to the above. This fixes it.


pages_builder/assets/javascripts/vendor/bind.js
```diff
@@ -0,0 +1,40 @@
...
+                 : oThis,
+                 aArgs.concat(Array.prototype.slice.call(arguments)));
+        };
+
+    fNOP.prototype = this.prototype;
+    fBound.prototype = new fNOP();
+
+    return fBound;
+  };
+}
```
carolinegreen: Might want to add a newline as we seem to be following this convention in most of the other toolkit javascript files, but apart from that 👍

allait: I think 4th wall doesn't recognise "👍" in the diff comments, only on the PR itself

carolinegreen: Yeah, it's a shame.

tombye: Good catch, it seems to be gone walkies when I copied it across.

PR #259, opened 2016-05-10
[#118094925] Remove duplicated question titles on one-page-per-question pages

pcraig3:
Doesn't look like much, but this brings in the important stuff from the [_single-question-page.scss](https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/master/app/assets/scss/_single-question-page.scss) rules that were before now stored exclusively in the supplier app.

Adds some more spacing between `.question-advice` and `.hint` text because the hint is more closely bound to the input field or some nonsense that ralph talks about sometimes. 

Look at the [side-by-side](http://stackoverflow.com/questions/24319505/how-can-one-display-images-side-by-side-in-a-github-readme-md) images down below and be amazed.

[>> Story on Pivotal](https://www.pivotaltracker.com/story/show/118094925)

***

### Screenshots

Before before            |  Now
:-------------------------:|:-------------------------:
screen shot 2016-05-10 at 18 06 45  |  screen shot 2016-05-10 at 18 07 01

Before (multiquestion)            |  Now (multiquestion: note the hint text is a bit lower)
:-------------------------:|:-------------------------:
screen shot 2016-05-10 at 18 17 17 | screen shot 2016-05-10 at 18 17 43


ralph-hawkins:
SICK

pcraig3:
:O=

toolkit/scss/shared_placeholders/_visuallyhidden.scss
```diff
@@ -0,0 +1,4 @@
+%visuallyhidden {
+  position: absolute;
+  left: -9999em;
```
TheDoubleK: Is there a reason for doing this rather than `display: none;`, which seems a lot less hacky.  What if the question is really long?  The tail end of it might end up on the screen if it's enough to fill 10000 pixels.  

If we have to do this hacky positioning out of screen for some reason, out on the right-hand side seems less risky? Anything over 4000 would do it at today's resolutions.

TheDoubleK: Is there a reason for doing this rather than `display: none;`, which seems a lot less hacky?  What if the question is really long?  The tail end of it might end up on the screen if it's enough to fill 10000 pixels.  

If we have to do this hacky positioning out of screen for some reason, out on the right-hand side seems less risky? Anything over 4000 would do it at today's resolutions.

TheDoubleK:
How does the extra space under the question advice look when (as with most questions) there is no hint text?

pcraig3:
It looks fine, Kev.

Before            |  Now
:-------------------------:|:-------------------------:
screen shot 2016-05-11 at 14 26 53 | screen shot 2016-05-11 at 14 27 16

toolkit/scss/shared_placeholders/_visuallyhidden.scss
pcraig3: It's a screen reader thing.  Screen readers ignore things that have `display: none;` styling.
https://www.gov.uk/service-manual/user-centred-design/resources/patterns/question-pages.html#accessibility

As far as putting it on the right or left, that seems like a reasonable point.  I was following [the existing style that was in the supplier app](https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/0578507db1d2cfb23b55f177f4d74f1de69bd3ce/app/assets/scss/_single-question-page.scss), but I would by curious to hear @tombye's opinion.

pcraig3:
👍

carolinegreen:
If this is generalised, an example in the toolkit would be useful.  Also what is visibility hidden being used for, and is it actually necessary?

toolkit/scss/shared_placeholders/_visuallyhidden.scss
tombye: The reason is to hide the content visually while keeping it available to users who don't use visuals to navigate the page. Here's an article explaining the technique: http://webaim.org/techniques/css/invisiblecontent/

www.gov.uk have styles for all the different use-cases around how to control the display of elements with CSS:

https://github.com/alphagov/govuk_template/blob/86106bfb332ff5c0346b92310a4117daeb2d8c8d/source/assets/stylesheets/_accessibility.scss

From memory, the only reason we position from the left is because standard page-flow is left-to-right but it's not a big problem if we decided to choose a different one instead.

Finally (!) the reason `em`s are used for the unit are because they scale with the text and 9999 `em`s are going to be a lot of pixels (at least 16 x 9999).

TheDoubleK:
I wonder... seeing that the things we are hiding are things that are otherwise duplicated on the page... would `display: none;` be better here, so people using accessibility tools don't have to have the same thing twice either?

carolinegreen:
I don't think it's clear without an example whether invisible labelling is necessary or not, as the advice with screenreaders is to avoid it where possible if the cue is already there for sighted users and not specifically a visual cue.  This is also mentioned in the article @tombye linked to.

tombye:
I agree this could do with a example to demonstrate how these styles are expected to be applied. I think the ambiguity around use of visually hidden could go away if this was there.

pcraig3:
So I added an example "Single question page" that's living under the "Forms" nested list.  It's the first one of those that's not using a template (it's just pure HTML stolen from the app), so it won't update itself as the question template evolves.
But it's pretty okay to explain the point of what we're doing.

If you're running the toolkit, you can see it at [http://localhost:8000/forms/single-question-page.html](http://localhost:8000/forms/single-question-page.html)

As to @TheDoubleK's question, my understanding was that:
- form inputs always need labels, and
- pages always need titles

Which is what's suggested by [the gov.uk guidance](https://www.gov.uk/service-manual/user-centred-design/resources/patterns/question-pages.html#accessibility).  

Seems to me they're both necessary.

carolinegreen:
I still wonder whether it how necessary the visually hidden legend is in this particular case.  It's one question, with one input box and doesn't appear to have visual indicators that make it a different experience for people who can see.

pcraig3:
I asked @tombye and his opinion was that all form inputs need `<label>`s or `<legend>`s because a screenreader will always read the contextual title before the value in the `<input>` field. 

So while it seems like it might be superfluous, but screen readers treat them differently enough that it's worth keeping both.

TheDoubleK:
That was harder work than you expected.  The whitespace looks lovely though.
:+1:

carolinegreen:
So I asked about it as well, and the reason for having hidden labels was to give extra context if someone left the screen open and came back later which at least makes more sense.

pcraig3:
<pre>whitespace                                              is                                                              the                                                                                                      best </pre>

PR #261, opened 2016-05-13
Add _reset.scss to toolkit

pcraig3:
All of our frontend apps have reset files, of which almost all of the CSS is identical.

`_reset.scss` in the
- [buyer app](https://github.com/alphagov/digitalmarketplace-buyer-frontend/blob/61c4a9be31f1a8564f670ec8821e2f22cccf9b29/app/assets/scss/_reset.scss)
- [supplier app](https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/9c35949cb5dd302125ce41662ea2383a77c5a6f2/app/assets/scss/_reset.scss)
- [admin app](https://github.com/alphagov/digitalmarketplace-admin-frontend/blob/5a4cabf8ee0364a51e897717068fd76e3af82707/app/assets/scss/_reset.scss)

Using the [`_forms.scss`](https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/0cf570f2528b0e94cb691a72d3258d460353ff08/app/assets/scss/_reset.scss) from the supplier app because it's the completest one.

This means I can remove a bit of styling from the `_lists.scss` file and import the reset instead.

### DIFF
--- a/toolkit/scss/shared_scss/_lists.scss
+++ b/toolkit/scss/shared_scss/_lists.scss
@@ -1,3 +1,4 @@
+@import "shared_scss/_reset.scss";
 @import "_measurements.scss";
 @import "_conditionals.scss";
 @import "_typography.scss";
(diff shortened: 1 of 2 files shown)


tombye:
Good idea 👍

PR #262, opened 2016-05-13
[What do we think?] Return labels in option-select boxes to size 16.

pcraig3:
_**Note:** this isn't a story, but it's something that changed unintentionally that we never really addressed.  If we want it back, now's the time._

***

### Smallen our option select labels

At some indefinite period in the history of our apps, our option select labels were embiggened to godzilla-like proportions.

This resets them to a modest, respectable size.

***

TOO BIG            |  nice
:-------------------------:|:-------------------------:
screen shot 2016-05-13 at 17 32 03  |  screen shot 2016-05-13 at 17 32 51


tombye:
I think good! 👍  (Though it needs a rebase.)

PR #264, opened 2016-05-18
[#18469059] Fix pricing fields

pcraig3:
Shrinking the negative margin fixes the problem almost entirely.

Bumping up the breakpoint fixes the rest of it.

Explanation below about the negative margins.
More detailed explanation in the commit messages.

***

### This is the size of our two-column div

screen shot 2016-05-18 at 12 12 21

### This is our pricing div without a margin applied
screen shot 2016-05-18 at 12 12 09

### This is our pricing div with `margin-right: 50%`
screen shot 2016-05-18 at 12 12 41

### This is our pricing div with `margin-right: -50%`
screen shot 2016-05-18 at 12 12 57

### This is our pricing div with `margin-right: -100%`
screen shot 2016-05-18 at 12 13 08

## Conclusion: `-100%` is too big.


tombye:
👍

PR #265, opened 2016-06-02
[#18469059] Fix pricing fields_final_FINAL_V2.docx

pcraig3:
Bumps up the breakpoint like I said I was doing in [the past pull request for this](https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/264) but didn't actually.

Using the previous breakpoint, there was a brief period where the layout would break between 950-900px wide (because the final pricing field would be dropped onto the line underneath).  

One this is merged, I'll update it in the various frontend apps.

*** 

### Means this no longer happens

screen shot 2016-06-02 at 10 21 07


tombye:
👍

PR #266, opened 2016-06-06
[#118469059] Bump up pricing breakpoint even further

pcraig3:
So, it turns out that our breakpoint works perfectly until it doesn't.

In our case, enforcing scroll bars on browser windows divorced the page width from the browser width, which dropped the last field down a line for very specific browser widths.

Bumping the breakpoint up 9 pixels to mitigate the problem as much as possible without going over `960px` (which is the max-width we use for our central column).

No problems if scroll bars aren't present, and very unlikely to be a problem otherwise.

***

## adding a scroll bar breaks the layout (for a very specific browser width)
price

### DIFF
--- a/toolkit/scss/forms/_pricing.scss
+++ b/toolkit/scss/forms/_pricing.scss
@@ -3,7 +3,7 @@
 @import "_shims.scss";
 @import "_typography.scss";
 
-@include media($min-width: 950px) {
+@include media($min-width: 959px) {
   .column-two-thirds .pricing {
     margin-right: -50%;
   }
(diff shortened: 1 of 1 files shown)


TheDoubleK:
How you make that nice animation? Not that I will ever take the time to do one, but it's great.

pcraig3:
[LICEcap](http://www.cockos.com/licecap/)!!!  A+++++ 5/5 would download again.

TheDoubleK:
:+1:

TheDoubleK:
thumbsup

pcraig3:
Thanks @TheDoubleK  !! 😄

pcraig3:
github

ralph-hawkins:
Yo dawg!

ralph-hawkins:
Yo dawg!!

ralph-hawkins:
Yo dawg!

PR #269, opened 2016-06-23
Extend `instruction-list` pattern

pcraig3:
As part of [the "Signing - new sign agreement flow" story](https://www.pivotaltracker.com/story/show/121577345), we've extended the instruction list to match @laurenceberry's design for starting to sign [a G-Cloud 8 framework agreement](http://framework-agreements.herokuapp.com/supplier/contract-start). 
<sub>I can DM you the password.</sub>

In order to have more flexibility when building our instruction list items, we removed the `extra` key in favour of `top` and `bottom` keys. This is a breaking change, but currently we're only using this instruction list in the supplier app, so it should be easy enough to fix.  The way that we're currently using instruction lists, we can just use `top` or `bottom` instead of `extra` and the rendered list will look exactly the same.
 
It's worth noting that the instruction list is currently quite a rigid pattern and if we find that we need to use it differently for something else, it should probably be exploded into configurable macros like we've done with service summary table cells.

[>> Story on Pivotal](https://www.pivotaltracker.com/story/show/121577345)

***

### Screenshot
image


TheDoubleK:
Not sure why this has got so old, sorry... looks fine to me :+1:

PR #270, opened 2016-07-05
Two thirds summary field

TheDoubleK:
Required for this story: https://www.pivotaltracker.com/story/show/121479497

Adds a `two_thirds` option to summary table rows.

The current `wide` option still works, but the underlying style has been re-named from `.summary-item-field-first-wider` to `.summary-item-field-first-half`.

This is a breaking change because the supplier app currently uses `.summary-item-field-first-wider` explicitly, so will need to be changed when the toolkit version is bumped there.

New example added looks like this:
screen shot 2016-07-05 at 12 32 27


ralph-hawkins:
👍 👍 👍

PR #274, opened 2016-10-06
Allow flag to external links for new window

Wynndow:
Added for [this story on Pivotal](https://www.pivotaltracker.com/story/show/128842457).

Previously external links would only open in the same window. This
allows a flag to be passed to be set to true which makes them open in a
new window.


pages_builder/pages/external-link.yml
```diff
@@ -4,3 +4,4 @@ examples:
   -
     link: https://www.example.com'
     text: Example
+    newwindow: true
```
idavidmcdonald: I'm not so familar with how we usually do things on the toolkit but one option would be to instead pass in a `target` attribute which is more flexible then just one option (`_blank). However, using `newwindow` abstracts away from the HTML which might be what we want with the toolkit? Not sure if either way is preferred but may be good to think about if you haven't already.

Oh and if we keep it as `newwindow` then it looks like convention is to use camel case (`newWindow`).

VERSION.txt
```diff
@@ -1 +1 @@
-17.0.0
+17.0.1
```
Wynndow: Yep, will change

pages_builder/pages/external-link.yml
Wynndow: Turns out it doesn't support markdown, but it does support html. I've updated link to actually be a link.

PR #275, opened 2016-10-17
Message headings are safe

TheDoubleK:
We need this to avoid:
screen shot 2016-10-17 at 11 04 51


Wynndow:
👍

idavidmcdonald:
👍

allait:
@TheDoubleK where is that `&nbsp;` coming from? I was really hoping we'd start to get rid of `| safe` filters.

TheDoubleK:
It's from the frameworks repo messages. I anticipated that @ralphhawkins would say that we shouldn't ever let the "2" of "Digital Outcomes and Specialists 
2" wrap round to a new line on its own.
So I added a bunch to the DOS2 content. We looked for ways round it last time and failed to find any...

allait:
@TheDoubleK I thought when we abandoned the spike for G-Cloud for this the outcome was that using a unicode nbsp character would be a better approach since it requires unicode safety (which is desirable anyway) and doesn't require `| safe`

TheDoubleK:
Ah OK, maybe? I don't remember that.  I'll put something in out backlog to sort it out, drop the safe and update the content everywhere.

TheDoubleK:
Made a bug to get fixed: https://www.pivotaltracker.com/story/show/132662099

PR #276, opened 2016-10-21
Link button with advice

Wynndow:
Part of [this story on Pivotal](https://www.pivotaltracker.com/story/show/129838641).

Link buttons were being rendered with a large margin on the top. This
was fine if there was no advice associated with the button, but meant
that the advice and the button looked ungrouped if there was.

By setting advice to true when using the template, it reduces the margin
to 15px and makes the advice and button look much more cohesive.

Before:
screen shot 2016-10-21 at 14 26 05

After:
screen shot 2016-10-21 at 14 20 09


TheDoubleK:
We need this exact same thing for green submit buttons for the new framework agreement signing flow. I'll follow the pattern here (pending approval from @laurenceberry)

idavidmcdonald:
Would be super nice if the example one was a bit clearer about what it actually does by including some text (i.e. advice) above it. At the moment the examples look like they do the same thing (spare the different class)

image

PR #278, opened 2016-10-24
Add heading to dmspeak

tombye:
Adds a 27px-size heading to the DM Speak styles and a placeholder to share these blocks of styles across different selectors.

## Why?

27px headings are currently only supported in summary tables patterns:

image

Pages like these make use of the style but use a mix of the [summary table heading styles](https://github.com/alphagov/digitalmarketplace-buyer-frontend/blob/master/app/templates/content/index-crown-hosting.html#L42) and app-specific CSS due to no other patterns being available in the toolkit.

- https://www.digitalmarketplace.service.gov.uk/terms-and-conditions
- https://www.digitalmarketplace.service.gov.uk/crown-hosting

## What was done

This introduces a placeholder that can be used in existing Sass and a new heading style in DM speak for long form copy. This should make the CSS available for all the existing use cases.

Note: also includes some fixes for a few issues around [Sass compilation](https://github.com/alphagov/digitalmarketplace-frontend-toolkit/commit/b17451d70c5bbdc5f1ea9ae85d52d76e7d7fc3a4) and [character encoding](https://github.com/alphagov/digitalmarketplace-frontend-toolkit/commit/1743d683b1b43c3fad2c6497d9a8b1f5435c31e7).

## Why now?

Chris had to add one of these headings as part of [this story on pivotal](https://www.pivotaltracker.com/story/show/129838641) (see PR: https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/276). He used the summary heading styles with the idea that once these styles were available in the toolkit they could replace that code.


ralph-hawkins:
👍

PR #282, opened 2016-10-26
Green button less space option

TheDoubleK:
We need to have a button closer to the text above it in some cases, e.g.:
# Existing "input.button-save" spacing
screen shot 2016-10-26 at 16 48 48

# New "input.button-save-with-advice" spacing
screen shot 2016-10-26 at 16 47 16


toolkit/templates/button.html
```diff
@@ -1 +1 @@
-<input type="submit" class="button-{{ type }}" {% if name %}name="{{ name }}"{% endif %} value="{{ label }}" />
+<input type="submit" class="button-{{ type }}{% if advice is defined %}-with-advice{% endif %}" {% if name %}name="{{ name }}"{% endif %} value="{{ label }}" />
```
allait: It turns out that `is defined` is not supported by Nunjucks, which makes it hard to reuse our templates in the prototype, so we're trying to phase it out (slowly). `{% if advice %}` does pretty much the same thing in this case, so I'd suggest we go with that.

allait:
Since this is a common thing: @tombye, @pcraig3, do you mind taking a look at this?

tombye:
Me and @TheDoubleK talked about this offline and this is the best solution though I should explain why.

The reduced vertical whitespace is to show a relationship between the button and the text. I suggested calling it a variant of `.button-save` using the [thing + type pattern](https://github.com/alphagov/digitalmarketplace-frontend-toolkit/blob/master/toolkit/scss/README.md#thing--type) with a type of `with-advice`.

The approach is similar to the one we took for the problem we had with link buttons (see https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/276) where we decided to use the existing pattern of the text above a control being information about its use, reflected in a class on the element parenting both which can effect their CSS.

We need to explain our use of vertical space better in future.

risicle:
Not totally sure about the `with_advice` naming, though generally seems sensible. :+1:

tombye:
The naming only appears in a few places so we can change it (outside of this PR).

Anything that explains the reason for the lack of top margin (that considers what the margin is for) is fine by me.

PR #283, opened 2016-10-27
Remove use of 'is defined' operator

tombye:
It's not needed here and causes problems when these macros are used with [nunjucks](https://mozilla.github.io/nunjucks/) (which doesn't support it and is the templating language our prototype uses).


allait:
:+1:

PR #284, opened 2016-11-01
New DM Speak styles

tombye:
Adds a new 19px heading style (from [govuk elements](http://govuk-elements.herokuapp.com/typography/#typography-headings)), used in pages like [find an individual specialist](https://www.digitalmarketplace.service.gov.uk/buyers/frameworks/digital-outcomes-and-specialists/requirements/digital-specialists), so all the headings used in long form content are represented.

dmspeak-new-heading

Also includes:
- spacing underneath lists in dmspeak so it matches paragraphs
- removes any spacing above the first heading in a dmspeak block

From a review of our pages, we use 48px and 36px for page headings (which are already supported in [that pattern](http://alphagov.github.io/digitalmarketplace-frontend-toolkit/page-heading.html)) and 27px and 19px for in-page content.


toolkit/scss/shared_placeholders/_dm-typography.scss
```diff
@@ -1,10 +1,30 @@
 @import "_typography.scss";
+@import "_measurements.scss";
 
-%heading-xmedium {
+// All heading styles, except heading-xmedium, are copied from GOVUK Elements as of this commit:
```
idavidmcdonald: This could be better please. What do we mean by 'copied'. Is it just in terms of copying class names etc.

heading-small is mentioned here and heading-xmedium. What about things like heading-large? Where is that defined? Is that defined? Why are these different?

tombye: I've made the comments specific to the `%heading-small` block and moved it to be above it to make it more obvious. Does that help?

pcraig3: Sorry to be pedantic, Tom, but _I_ think it would better to also have another comment describing where the other header styles are defined -- sort of like the one was there before.

Maybe something like:
```css
// `heading-xlarge`, `heading-large`, `heading-medium`, and `heading-small` are inherited from GOVUK Elements as of this commit:
// https://github.com/alphagov/govuk_elements/commit/7aaa646a59019b6a90e8a88ddc81ac61950b9efe
```

(or whatever you think would be better).  

Personally, I didn't make the connection that the other headings were defined already in until  I read the other comment you deleted.

Should be good to go after that :)

pcraig3:
👍  once conflicts are resolved!

PR #285, opened 2016-11-03
Make phase-banner into real pattern that we might actually use.

pcraig3:
The [buyer](https://github.com/alphagov/digitalmarketplace-buyer-frontend/blob/master/app/templates/_phase_banner.html) and [supplier](https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/master/app/templates/_phase_banner.html) app are both using their own versions of the phase banner, which are nearly identical.

Talked with @tombye about this yesterday and he said that maybe the thing to do was make the phase into a variable.

Seems to me there's 3 ways forward.

1. Remove `beta-banner` pattern: we don't need it, just let the frontend apps do their thing.
2. Make `beta-banner` entirely static: both apps are just using the same markup and we don't anticipate a need to change either of them.
3. Make the `phase` into a variable: this way we have a pattern, even though for all [intensive porpoises](http://cdn.styleforum.net/7/7a/7a50d724_a-man-with-a-porpoise-25171-1260418770-21.jpeg), we don't really need the flexibility. 

I think we want **2** or **3**.  What do you think, @tombye?


ralph-hawkins:
Don't forget the version on the homepage that has white on blue.

pcraig3:
@ralph-hawkins I was thinking that would be better handled by the app itself, but maybe you're right that it would be better handled by the toolkit.

ralph-hawkins:
Having all the styles in the same place sounds good to me

pcraig3:
Sorry, @ralph-hawkins, but @tombye has vetoed the frontpage changes.  Keep the styles in the supplier app because the homepage is a special thing.  

:boom: :boom: :ralph: :boom: :boom:

tombye:
I did, sorry Ralph. I just think the home page styles are specific to that cascade so it feels a bit strange viewing them out of that context.
At the least, this forms a good base for 99% of usage across the apps so it can be grown if needed (we can chat about it and see).

tombye:
@pcraig3 it seems like renaming the macro file makes this a breaking change, shouldn't it be `18.0.0`?

pcraig3:
ughhhh.

fiiiiiiine.

tombye:
Once all conflicts are sorted 👍

PR #286, opened 2016-11-04
Remove `safe` and `markdown` filters from templates

pcraig3:
So it turns out that using `safe` and `markdown` (which is more permissive than `safe`) isn't actually that safe, [despite the deceptive name](https://github.com/OWASP/railsgoat/blob/master/app/views/layouts/shared/_header.html.erb#L27-L29). 

This pull request removes all uses of `safe` and `markdown` from our templates, as well as removes two outdated patterns that we have stopped using (namely, [`search-results`](http://alphagov.github.io/digitalmarketplace-frontend-toolkit/search-results.html) and [`framework-notice`](http://alphagov.github.io/digitalmarketplace-frontend-toolkit/framework-notice.html)).

It's a pretty large breaking change because in quite a few places we're passing stuff into the templates that we expect will be escaped, but @allait and I are taking care of pulling them into the frontend apps. 

Markdown formatted text in the content is still allowed, it's just going to be handled in the content loader now instead of here, so some of our examples have had to change to just using straight HTML rather than asterisks and dashes.


allait:
👍

idavidmcdonald:
Would be good to add a record to the changelog for the breaking version bump please :)

PR #287, opened 2016-11-10
Use string concatenation instead of format method

tombye:
We use our macros with the nunjucks templating
language, written for JavaScript. Use of the
`format` method is causing an error in this
context because it is not supported in JavaScript.

This switches use of `format` out for string
concatenation with the `+` operator, something
supported by both Python and JavaScript.


TheDoubleK:
Where do we use nunjucks?  Is that a new thing?

allait:
@TheDoubleK we use it in the prototype.

👍

PR #289, opened 2016-11-16
Add show-hide capacity to the frontend toolkit

pcraig3:
### Support shown/hidden fields

Adds the ability for a boolean question to show or
hide 'text' or 'textarea' input fields.
[Docs for usage are here](https://github.com/alphagov/govuk_frontend_toolkit/blob/master/docs/javascript.md#showhide-content).

It's worth pointing out that the scope of this is
limited to these particular input types because
these are how we plan to do 'nice-to-have'
questions. It could be expanded to support other
fields if needed, but for now, these changes aren't
really general things we need to do.

Also worth pointing out that the `hidden` attribute
will create a `js-hidden` class which will only
truely hide the element if javascript is enabled.

### Bumped the govuk_frontend_toolkit version 

This is the version that actually contains the
show/hide javascript.  It's a bit pointless updating
this because our toolkit doesn't actually use this
(the frontend apps are responsible for initiating
the javascript), but I've just updated it here for
completeness.

I thought about doing an example of this being used,
but it seems like it doesn't really work the way
oour other examples do (it includes two different
types of input fields), and it's also not really a
feature that we've built.  Instead, the example to
check out should be [the one in the govuk frontend
toolkit](https://github.com/alphagov/govuk_frontend_toolkit/blob/master/docs/javascript.md#showhide-content).


allait:
👍 

Let's bump the minor version though.

tombye:
This is so good it deserves to be featured in the documentation. On the [selection buttons page](http://alphagov.github.io/digitalmarketplace-frontend-toolkit/forms/selection-buttons.html) perhaps?

Also, @allait, isn't this introducing a feature and so `19.1.0` works?

tombye:
I've added an example to the docs pages:

image

This required some hacking in the code that generates the pages but nothing that should break anything. Should be good to merged now.

pages_builder/generate_pages.py
```diff
@@ -161,8 +161,15 @@ def render_page(self, root, file):
                 for index, example in enumerate(partial["examples"]):
                     grid = partial.get('grid')
                     if isinstance(example, dict):
+                        # if the example has some html it needs to be displayed, cache it and remove from the parameters example
+                        surrounding_html = example.get('surrounding_html', None)
+                        if surrounding_html:
+                            del example['surrounding_html']
```
pcraig3: Bit late now, but you could do a `.pop('surrounding_html', None)` and not have to do a bunch of this.

PR #290, opened 2016-11-23
Single field pricing question

Wynndow:
If the case is that a pricing question only has one field, we should
render a simpler label.

Also, there's no need to use pricing columns and the more complicated
code that comes with it. We use the simple text box format instead.


pcraig3:
🙀

Wynndow:
@pcraig3 Lunatic

tombye:
At some point we need to make it so the top pattern is just an import of the `forms/textbox.html` macro. I mean so any changes to the textbox this code is copied from update automatically.
Doing this now would mean changing the import path used across the apps so for now, this is 👍

Wynndow:
@tombye Hey Tom, we spotted something after you gave this a thumbs up. Could you take a look at that last commit for us?

tombye:
Good spot. Now needs a rebase + the version bump changed to be to the new version but once that's done: 👍

PR #291, opened 2016-11-25
[#133024283]summary table: add 16px left padding to .summary-item-field-with-action so that action text doesn't run together when tightly constrained

risicle:
For story https://www.pivotaltracker.com/story/show/133024283

We're targeting the service listing table here.

Before/Without horizontal padding:
Without horizontal padding

After/With horizontal padding:
With horizontal padding


TheDoubleK:
I guess the question is whether the `field-with-action` will always be at the far right-hand side of the table?  If so then this works fine, but if not then might it be better to add 6px padding on each side of the base `summary-item-field` class?

I'm not sure if that might cause any new issues, but seems like it should be OK?

Probably worth getting a quick opinion from @tombye.

risicle:
Well my reasoning was that since it's `test-align: right` then it's always going to have rightwards "gravity". I actually wonder if 16px padding is too much horizontally as it's such a precious resource and does look ok with less.

TheDoubleK:
Ah yes good point about the inherent "right-ness" of it. 12px seems to be what we use elsewhere, so this is probably right in terms of keeping things looking balanced.
:+1:

risicle:
Yeah still like to hear @tombye 's final word on this

tombye:
I've had a chat with @risicle about this. In general we could do with looking at tables for smaller screen sizes but for now this'll ensure there's enough clearance for clicking on those links.

👍

PR #292, opened 2016-11-28
Add a npm install build step to docs generation

tombye:
The part of the script that makes the JS for the docs site was rewritten to copy all the `node_modules` JS across in this commit: https://github.com/alphagov/digitalmarketplace-frontend-toolkit/commit/9e32356eb296f798674b6c8c4ee117389da07a60#diff-76a5ddb47815102dec7ecd599047c69bR227

Only problem is there's no `npm install` step to the current docs generation (and so no `node_modules` folder) which meant the script was breaking. This adds it so should fix it.


allait:
👍

PR #293, opened 2016-11-29
Base build on HEAD SHA rather than latest tag

tombye:
Sometimes we want to make changes to the documentation code which shouldn't need a version
bump. This changes the build to be based on the latest HEAD commit on the master branch rather than the latest tag.


allait:
👍

PR #294, opened 2016-11-30
Fix how usage is called

tombye:
Fixes this problem with https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/293.


pcraig3:
👍

PR #295, opened 2016-12-22
Make module loader recallable

tombye:
The module loader usually sits at the end of each app's `application.js` and initialises all modules added to the `GOVUK.GDM` namespace up until that point:

The order of `application.js` is roughly:

1. load all base libraries, ie analytics and jquery
2. load all modules from this toolkit and attach them to the `GOVUK.GDM` namespace
3. run the `module-loader.js` code which looks for any modules in `GOVUK.GDM` and initialises them

This means if you want to load any more JavaScript files into the page you can't run them because they load after `application.js`, at which point `module-loader.js` has run.

This pull request refactors `module-loader.js` so you can run it more than once without re-initialising any modules already called.

The use case for this was some code @risicle was writing to make the stats we have as part of the admin app public. In this case, that view needed the [d3](https://d3js.org/) library but it was the only view that did. That code never made it to production but I think there's some value to this and the new form doesn't break any existing uses.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -6,21 +6,24 @@ var environment;
 var repoRoot = __dirname + '/';
 
 gulp.task('test', function () {
-  var manifest = require(repoRoot + 'spec/javascripts/manifest.js').manifest;
+  var manifest = require(repoRoot + 'spec/javascripts/manifest.js').manifest,
+      fixPaths;
 
-  manifest.support = manifest.support.map(function (val) {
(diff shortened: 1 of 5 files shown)


tombye:
While this could be a useful change in future, it requires someone's time to review so I'm going to close it for now.

If the need arises in future for our pages to call any scripts apart from `application.js`, this should be re-considered. It's fairly isolated anyway so shouldn't go stale in the meantime.

PR #298, opened 2017-01-24
Fix GOV.UK style import overriding display type

TheDoubleK:
This is causing an issue with really long underlines in production.  Agreed with Ralph that we don't need the underlines, which means we can drop the `inline` on the header, which was a weird thing to do.

## BEFORE
screen shot 2017-01-24 at 16 42 31

## AFTER

**NOTE: UPDATED NOW WE'RE DROPPING THE UNDERLINE**
screen shot 2017-01-25 at 10 28 16


pcraig3:
👍  Don't know how it happened either, but the fix works.

tombye:
I did a bit of detective work on this.

The browse-list pattern has custom underlines which are actually fakes made by using a border-bottom.

Whereas normal underlines only stretch to the end of their text, using a border-bottom means the custom ones stretch the length of the HTML element containing the text. Long story short, the `heading-24` mixin which styles the browse-list text was changed to `display: block`, meaning it now occupied all the space of its containing element (which the border, ahem, borders).

The custom underlines were added on Jul 10, 2015:

https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/101/files

The `heading-24` mixin was made block level on Jun 20, 2016:

https://github.com/alphagov/govuk_frontend_toolkit/pull/200

This was added to the GOVUK Frontend Toolkit in version 4.13.0

https://github.com/alphagov/govuk_frontend_toolkit/commit/3703690bc3e6db003dd3eb8d7b15a65273faec01

The Digital Marketplace Toolkit docs got this version on Jan 9, 2017:

https://github.com/alphagov/digitalmarketplace-frontend-toolkit/commit/74b865d170471433f7ee09155fb73ebfcf2ac76e

The supplier app got it on Jan 5, 2017:

https://github.com/alphagov/digitalmarketplace-supplier-frontend/commit/5c88a2a9bd52fcbf695407ef22340f7fc0ce6179#diff-b9cfc7f2cdf78a7f4b91a753d10865a2

All the other apps will get it when they bump their version of the GOVUK Frontend Toolkit.

This change may fix it but we don't really want headings to be `display: inline` as any text following them should be on the next line. @ralph-hawkins can we just lose the custom underlines?

galund:
Incidentally we have a ticket about updating this lib I think? (#138233889)

TheDoubleK:
OK, will leave merging this until we can decide the best way forward in the morning.  But it's an issue currently visible in production so would be nice to get it out sooner rather than later.  Like tomorrow morning.

pcraig3:
Nice work @tombye! I remember something about custom underlines going in but I wasn't too clear on what exactly they were doing. 

Turns out there's a bunch of different ideas about how you do custom underlines if you're interested in the topic. :/

pcraig3:
I was going to question removing the underline altogether, but it looks like that's normally what we do anyway.

image

Thumbs up from me, although it feels like @tombye should have the last word.

toolkit/scss/_browse-list.scss
```diff
@@ -16,17 +16,13 @@
 .browse-list-item-link,
 a.browse-list-item-link {
 
-  display: inline;
   @include heading-24;
   font-weight: bold;
   text-decoration: none;
-  border-bottom: 1px solid; // No border colour, is inherited from text colour
   position: relative;
```
tombye: I don't think it needs either `position: relative` or `top: -12px` as they are

tombye:
There were still some styles left over and a `<span>` used only for the custom underline styling. I took them out (in https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/298/commits/926d1e37f82b358a29a9317ab8bb939f597d600e) so this should be good to go.

pcraig3:
Works, tightens the spacing, and reduces the complexity.  👍

TheDoubleK:
Nice one @tombye!  I've tested locally with this version and all looks good still.

PR #300, opened 2017-02-07
Use simple threaded server instead of `SimpleHTTPServer`

pcraig3:
To run the frontend toolkit thus far, we've been using a 
`SimpleHTTPServer`, which is limited by the fact that it can't 
serve simultaneous requests.

In our day-to-day usage, this hasn't been a huge problem, but 
more recently we've tried running Wraith against it, which it 
can't really handle. Even having a local toolkit page open in a
browser while running Wraith would cause the server to hang/fail.

To mitigate this, I created a small threaded version of the
SimpleHTTPServer based on a bunch of stackoverflow answers that
were all kind of saying the same thing.

Usage is exactly the same -- it just handles concurrent requests now. 

***

Helpful Stackoverflow answers:
1 - [Processing Simultaneous/Asynchronous Requests with Python BaseHTTPServer](http://stackoverflow.com/questions/12650238/processing-simultaneous-asynchronous-requests-with-python-basehttpserver)
2 - [BasicHTTPServer, SimpleHTTPServer and concurrency](http://stackoverflow.com/questions/2455606/basichttpserver-simplehttpserver-and-concurrency)
3 - [Multithreaded web server in python](http://stackoverflow.com/questions/14088294/multithreaded-web-server-in-python)


Wynndow:
pretty fancy...

pcraig3:
Thanks, @Wynndow!  I was actually trying to use parallelism in real life to i) improve the toolkit and ii) impress you, so things are looking pretty good right now.

Wynndow:
Clearly you missed the sarcasm of my first comment.

Wynndow:
Seriously though, pretty fancy...

tombye:
👍

PR #301, opened 2017-02-07
Copy govuk_frontend_toolkit into /pages_builder as a build step

pcraig3:
Previously, updating the `govuk_frontend_toolkit` meant that the
stuff in the `/node_modules` was updated but none of the code used
to build the frontend toolkit was. Copying the new stuff was left
as an exercise for the student.

Now I've added an extra bit to the build step ([inspired by the supplier frontend](https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/134c901a702fca65225558c6ba47ef03db763c12/gulpfile.js#L140)) which copies the stuff we want.

Means we can delete the contents of the `/pages_builder/govuk_frontend_toolkit` folder.


tombye:
👍

pcraig3:
Yeah, but it's not really a backwards compatible bugfix either.

Let's just leave it as is.

pcraig3:
(he says now that it's merged.)

PR #302, opened 2017-02-07
Bump toolkit and bring in new radios/checkboxes

pcraig3:
This pull request:

- bumps the `govuk_frontend_toolkit`
- bumps the `govuk_template`, removes external-link styles
- brings in styles for the new radios and checkboxes first popularised by @robinwhittleton of [govuk_elements](https://github.com/alphagov/govuk_elements/pull/296) fame
- adds 'role="button"` to links that are buttons
- bolds question headings

### new radios and checkboxes

New radios and checkboxes look nice, and I've removed most of our custom spacing rules as those in `govuk_elements` seem to work better with the new elements.
We also had a problem passing values to prefill our inputs with but that's been fixed.

### external links

Removing our external-link styles to keep consistent with the govuk toolkit, but we're keeping the macro because our apps still rely on the markup. 

### phase banners

Stopped passing a `phase` to the `phase-banner()` mixin.  Banners are now blue, but no visible changes to sizing or spacing.

### bold question headings

This is what [govuk_elements are doing](https://govuk-elements.herokuapp.com/form-elements/#form-checkboxes) and now it's what we are doing. 

Worth mentioning that I ran [wraith](https://github.com/BBC-News/wraith) against all the pages after updating the govuk_frontend toolkit and then only real visual changes I found were to do with external links and the beta banners. 

***

# gifs

| old | new |
|-----|-------------|
| rads1 | r2 |

# screenshots

- **old**: what the selection buttons looked looked like before
- **old spacing**: what the new elements looked like with digitalmarketplace-specific spacing preserved
- **new spacing**: what the new elements look like with our hyper-rationalised vertical spacing

The differences between the old spacing and the new spacing are:
- form elements are more tightly grouped
- the labels have been moved away from the input elements (this is particularly noticeable when the focus border is applied) 
- the spacing between all meta-elements (question_advice, hints, optional tags, validation) and the questions themselves is now equal

## radios

| old | old spacing | new spacing |
|-----|-------------|-------------|
| screen shot 2017-02-10 at 12 14 01  | screen shot 2017-02-10 at 12 14 33  | screen shot 2017-02-16 at 11 00 46 |

## checkboxes with hint and question advice

| old | old spacing | new spacing |
|-----|-------------|-------------|
| screen shot 2017-02-16 at 10 29 36  | screen shot 2017-02-16 at 10 35 29   | screen shot 2017-02-16 at 10 11 59   |

## new spacing 

All distances are now equal. 

screen shot 2017-02-16 at 10 11 59 copy


pages_builder/assets/javascripts/onready.js
```diff
@@ -15,10 +15,7 @@
   }
 
   if (GOVUK.SelectionButtons) {
-    var selectionButtons = new GOVUK.SelectionButtons('label.selection-button', {
-      'focusedClass' : 'selection-button-focused',
-      'selectedClass' : 'selection-button-selected'
-    });
+    var selectionButtons = new GOVUK.SelectionButtons('label.selection-button');
```
robinwhittleton: Usually the selection button JS [expects to apply to the input](https://github.com/alphagov/govuk_frontend_toolkit/blob/master/docs/javascript.md#selection-buttons), not the label. Is this working correctly?

toolkit/templates/external-link.html
```diff
@@ -1,3 +1,7 @@
 <div class="external-link-default">
-  <a href="{{ link }}" rel="external" {% if target %}target="{{ target }}"{% endif %}>{{ text }}</a>
+  <a
+    href="{{ link }}"
+    rel="external{% if target=='_blank' %} noopener noreferrer{% endif %}"
```
robinwhittleton: 👌

toolkit/scss/forms/_questions.scss
```diff
@@ -12,6 +12,11 @@
   .hint {
     margin-bottom: 0;
   }
+
+  // This preserves the vertical margins set on the .question-headings
+  legend {
```
tombye: Can we check this with @ralph-hawkins? This will add 10px to the spacing between legend and first field so we should make sure we want this.

This doesn't mean I'm against this exact change, it's just worth asking if we need to also adjust the padding and line-height so the behaviour of the CSS is correct but the visuals stay the same.

pcraig3: Yeah, sure. Good point.

The bottom margin is already there but it's getting ignored, which's means that we should either
i) add this so that it shows up, or
ii) remove the margin

Having the rule and not using it seems like the wrong thing to be doing.

toolkit/scss/forms/_selection-buttons.scss
```diff
@@ -36,39 +38,122 @@
...
+    clear: none;
+    margin-right: $gutter-half;
+  }
 }
 
-// Remove focus from radio/checkboxes
-.js-enabled .selection-button-focused input:focus {
-  outline: none;
+// To stack horizontally, use .inline on parent, to sit block labels next to each other
+.inline .block-label {
```
pcraig3: Yup, makes sense.
Lose terseness but gain clarity. 

Read [a blog post by Harry Roberts](https://csswizardry.com/2017/02/code-smells-in-css-revisited/) the other day and that seems to be one of his ideas.

toolkit/scss/forms/_questions.scss
```diff
@@ -1,16 +1,35 @@
-@import "_colours";
-@import "_typography.scss";
+@import "colours";
+@import "typography";
 @import "shared_scss/_dmspeak.scss";
 @import "shared_placeholders/_visuallyhidden.scss";
 
+
+// Spacing rules based on those found in govuk elements selection buttons
```
tombye: It might be good to move this into its own placeholder file, just incase we end up wanting to import `_questions.scss` into anything in future (in which case the selectors produced will appear in all those places).

toolkit/templates/forms/selection-buttons.html
```diff
@@ -8,9 +8,9 @@
 {% if error %}
   <div class="validation-wrapper">
 {% endif %}
-  <fieldset class="question first-question" id="{{ name }}">
+  <fieldset class="question first-question{% if inline or type=='boolean' %} inline{% endif %}" id="{{ name }}">
     <legend>
-      <span class="question-heading{% if hint or question_advice %} question-heading-with-hint{% endif %}">
+      <span class="question-heading{% if hint_underneath %} question-heading-with-hint-underneath{% endif %}">
```
tombye: Does the `question-heading-with-hint-underneath class exist? I can't find it in this repo (or this pull request).

toolkit/templates/forms/selection-buttons.html
```diff
@@ -40,51 +40,57 @@
...
           id="input-{{ name }}-yes"
           value="true"
           {% if question_advice or hint %}aria-describedby="{{ [question_advice_id, answer_advice_id]|join(' ') }}"{% endif %}
-          {% if value is sameas true %} checked{% endif %}
+          {% if default_yes %}checked="checked"{% endif %}
         />
       </label>
-      <label class="selection-button selection-button-boolean" for="input-{{ name }}-no">
+      {% set default_no = true if value is sameas false %}
+      <label class="selection-button selection-button-radio {% if default_no %} selected{% endif %}" for="input-{{ name }}-no">
```
tombye: This doesn't need a space before the `{% if default_no %}` opening block.

toolkit/templates/forms/textbox.html
```diff
@@ -28,7 +28,7 @@
       {% endif %}
     </label>
     {% if optional %}
-      <span class="question-optional" style="margin-top: 10px;">
```
tombye: Nice bit of tidying up this.

toolkit/scss/forms/_questions.scss
tombye: For the record, this was resolved by @pcraig3 working with @ralph-hawkins to make the vertical spacing for all questions consistent, whatever elements are present about the field.

toolkit/templates/forms/selection-buttons.html
tombye: The work done on vertical spacing (mentioned elsewhere on this PR) removed the need for `question-headings-with-hint`.

tombye:
Once it's good to be merged, 👍

PR #304, opened 2017-02-16
Follow up questions

allait:
### Remove 'is defined' check from textbox form template

`hidden` question property should only ever be set to `True` if present
(since the only thing it controls is whether the question should be
initially hidden), so we can remove the separate `if hidden is defined`
check, which would cause issues with Nunjucks.

 ### Add followup support for radio and checkbox questions

Merges the boolean and radio question form templates into one and
adds support for the multiple followup questions using the new
`Question.values_followup` property passed in as `followup`.


samuelhwilliams:
I think this is missing `{% if hidden %} js-hidden related-information{% endif %}` from the fieldset in list-entry.html.

allait:
@samuelhwilliams added hidden flag for list-entry questions
@tombye added some examples to combinations.html
@pcraig3 I've rebased this on top of current master and had to make some more changes to the selection-buttons.html, would appreciate an additional review to check that I haven't missed anything.

pages_builder/pages/forms/selection-buttons.yml
```diff
@@ -252,7 +252,9 @@ examples:
     question: Can you do this essential requirement
     question_number: 93
     value: true
-    followup: requirement-evidence-110
+    followup:
+      true:
+        - requirement-evidence-110
```
samuelhwilliams: This should be the other way around, shouldn't it?

```followup:
  requirement-evidence-110:
    - true```

Comment (allait) [inline pages_builder/pages/forms/selection-buttons.yml]: This is not YAML format used by the content loader, it's the data we'd normally pass into toolkit form macros, so this is what `.values_followup` would return, which is a reversed mapping.

Comment (samuelhwilliams) [inline pages_builder/pages/forms/selection-buttons.yml]: ah ok

File: toolkit/templates/forms/list-entry.html [hunk 2]
```diff
@@ -13,7 +13,7 @@
 {% elif message %}
   <div class="message-wrapper">
 {% endif %}
-  <fieldset class="question {% if first_question %}first-question{% endif %}" id="{{ id }}">
+  <fieldset class="question {% if first_question %}first-question{% endif %}{% if hidden %} js-hidden related-information{% endif %}" id="{{ id }}">
```
pcraig3: This is a thing we've done in a few places where we add extra spaces to classnames. 
If you remove the space after `class="question ` and put it before `first-question`, it's more correct.

toolkit/templates/forms/selection-buttons.html
```diff
@@ -45,67 +45,41 @@
...
+    {% endif %}
+
+    {% if type in ["radio", "boolean"] %}
+      {% set input_type = "radio" %}
+    {% else %}
+      {% set input_type = "checkbox" %}
+    {% endif %}
+
+    {% for option in options %}
+      {% set input_value = option.value if 'value' in option else option.label %}
```
pcraig3: Good call fixing this stuff.  I know @benvand was asking about it.

pcraig3:
I think go for it. 👍

PR #308, opened 2017-02-24
Update changelog

samuelhwilliams:
Some tests were broken by the questions-with-dependencies PR. Small update just to include an example migration required for the tests.


pcraig3:
This is a good catch. A+

👍

PR #309, opened 2017-02-24
Allow custom id prefixes to selection buttons

pcraig3:
Our new checkbox_tree code (which is not yet in the frontend toolkit) reuses the `selection-buttons.html` macro to generate lists of checkboxes.

This means that, as it stands, we end up with repeated ids for our checkboxes and radios if we call this macro multiple times.

This commit allows passing in a custom id prefix to our selection button inputs to which we then add a loop counter.


tombye:
👍

PR #312, opened 2017-02-28
Show hide chained followups

allait:
Since we're using the disclosure logic for chaining questions, there's a problem when a question that controls the visibility of another input is itself hidden.

So for three questions A, B and C, where A controls the visibility of B and B controls the visibility of C, if all of them are visible on the page at some point and then question A is answered with a value that hides question B, question C will remain visible.

To avoid this, we add the logic to recursively hide/show all controlled content if the controlling input is hidden/revealed.

followups


allait:
@pcraig3 it does now. In a separate commit, to make it easier to review.

pcraig3:
👍  after a rebase.

PR #313, opened 2017-03-02
Add checkbox tree pattern to toolkit

pcraig3:
This pull request

- adds the checkbox-tree pattern (to be used for selecting [service categories for G9](https://github.com/alphagov/digitalmarketplace-frameworks/pull/387))
    - this includes two versions of the full-page render and adding a new summary-table macro
- moves the code for rendering the `<input>` element from out of the `selection-buttons` macro
- fixes a small bug around possibly having unwanted whitespace in aria controls for inputs 
- fixes a bug where `id`s for selection-button would be incorrectly generated if an id wasn't actually passed in

### checkbox tree

The checkbox tree pattern has broken some of our assumptions about form components in the toolkit.  Some of the differences are:
- it will create two columns inside of itself -- other components don't do this
- in the left-most column, we need a “counter” component. This could be isolated as well, but the HTML is pretty straightforward and it’s only used in this pattern, so there seemed to be little value in putting the counter into its own file
- in the right-most column, we need to create a series of lists of checkboxes — effectively looping through an existing macro.  This is not something we’ve done in the toolkit before and thus the solution we’ve used elsewhere for importing other files is pretty graceless.

Also, since it needs to be full-width, it does what the pricing module does and uses a negative width to break out of its `.column-two-thirds` container.  Not unprecendented, but certainly a problem if we have to keep doing this. 

However, it still needs to live inside a form, it’s still one-question-per-page, and it still submits data the same way as any other element.  I don’t think it belongs anywhere else.

#### importing other macros

The worst thing in this pull request is [the `if` statement around importing other files within the toolkit](https://github.com/alphagov/digitalmarketplace-frontend-toolkit/commit/3bc1c4332b3ad678df76f2fbb66d3cc96a4bb8c8).  Even though all the relative file paths are the same, the ‘template root path’ is different in our apps vs the frontend-toolkit pages generator. The `if` statement only works because we control all of the apps consuming the toolkit and we know how their directory structures work.  This seems like a high-priority thing to fix because it’s super brittle and really ugly.

#### find matching items in two different arrays

The second worst thing about this pull request is [the logic where we want to open a checkbox list by default if one of its values has been checked](https://github.com/alphagov/digitalmarketplace-frontend-toolkit/commit/bc1dec9b1934d89f80382d7a22ec06e5b1685eb6).  The operation we need to do is “find matching items in two different arrays”, but I couldn’t think of a cleaner way to do this within jinja.  The argument I found online seems to be “don’t do this in your template”, but, due to the way we pass data to our templates, there didn’t seem to be an easier way to do this. Suggestions or better algorithm proposals welcome.

#### selecting stuff in the checkbox tree
checkbox_tree

####  summary table layout
screen shot 2017-03-02 at 11 42 13


TheDoubleK:
Are you planning to do the TODO bits here at all?

pages_builder/pages/forms/checkbox-tree.yml
```diff
@@ -0,0 +1,176 @@
...
+
+    {%
+      with
+      name="serviceCategories-10",
+      question="Which categories does your service fit under?",
+      question_advice="",
+      hint="You can choose multiple categories.",
+      id="serviceCategories-10",
+      options=[
+        {"label": "Archiving, backup and disaster recovery"},
```
pcraig3: Yeah, I'm just using the real categories we're using. 
It's not fun, but it was easy.

toolkit/templates/forms/_checkbox-tree-selection-buttons.html
```diff
@@ -0,0 +1,71 @@
+<fieldset class="checkbox-tree__selection-buttons">
+  <legend>
+    <h2>{{ name }}</h2>
+  </legend>
+  {% for supercategory in options %}
+    {% set id_prefix = 'input-{}-{}'.format(name, supercategory.label|replace(" ", "")) %}
+    {% if supercategory.options %}
+      <details
+          {% for _o in supercategory.options %}
```
pcraig3: It _is_ a bit odd but it's meant to be pretty self-contained and not clash with variable declarations elsewhere.

PR #316, opened 2017-03-06
Small updates to `checkbox_tree` macro

pcraig3:
This pull request contains a bunch of small updates to the `checkbox_tree` macro that I made while working on the JavaScript part of this for the supplier app.  While the added JavaScript isn't actually moving into the frontend toolkit at this point, these changes are generally just small improvements to the pattern.

Meaningful changes in this pull request are as follows:
- makes the counter element an "aria live" region: notify screen readers that the counter element is being updated 
- stops using the parent category name as part of the id: ids for groups of inputs will now look like `serviceCategories-1` rather than `serviceCategories-Accountingandfinance`
- Removes extra `input-` string from input ids: ids for individual inputs will now look like `input-serviceCategories-1-1` rather than `input-input-serviceCategories-Accountingandfinance-1`
- Adds a new summary table layout for checkbox_trees without children (it has bullets)
- Removes bottom margin from final list of checkboxes in checkbox_tree


TheDoubleK:
Isn't it helpful to have the category name as part of the input?

toolkit/templates/forms/_checkbox-tree-selection-buttons.html
```diff
@@ -1,30 +1,30 @@
-<fieldset class="checkbox-tree__selection-buttons">
+<fieldset class="checkbox-tree__inputs" id="checkbox-tree__inputs">
```
risicle: Setting the `id` like this does mean you're not going to be able to use the macro more than once on a page (and have it still be compliant)

pcraig3:
@TheDoubleK Not really. 

We're never actually using the input id for anything -- it's only the category-picker javascript that cares about it. 

While there probably _is_ some abstract value in having a human-understandable id, the problem I ran into was that the javascript falls over when the id contains invalid characters ([ie, brackets](https://github.com/alphagov/digitalmarketplace-frameworks/blob/810723c134fb491868751c5f0f7854772bf0604a/frameworks/g-cloud-9/questions/services/serviceCategoriesSoftware.yml#L153)).  I was using a string replace method to collapse all whitespace, which already felt a bit dicey, so once this problem appeared, I just gave up.  There doesn't seem to be a jinja method at hand to remove all offending characters and I don't want to be trying to make one either.

toolkit/templates/forms/_checkbox-tree-selection-buttons.html
pcraig3: Yes, that's what this means.  It's a pretty short-term fix.

pcraig3: The problem is that the javascript that runs alongside this has been making the assumption that there's only one category-picker on a page, and, for G9, that's all there will be. 

The example page in the toolkit has a bunch of them on a page, so that's where this becomes a problem.  

Right now we need the javascript working in the supplier app against the real-live category-picker more than we need the javascript to work in the frontend toolkit example page.

risicle: Acknowleged.

toolkit/scss/forms/_checkbox-tree.scss
```diff
@@ -5,7 +5,7 @@
 
 @include fake-full-width('.checkbox-tree');
 
-fieldset.checkbox-tree__selection-buttons > legend {
+fieldset.checkbox-tree__inputs > legend {
   @extend %visuallyhidden;
```
carolinegreen: I can see it's not a change in this request, but I'm not sure you should be using extends here instead of a mixin, as it doesn't really make sense as an extension.  I think this should probably go into tech debt though if you are trying to get things out quickly.  I'm sure we were talking about this when I was still on Digital Marketplace.

pcraig3: Yeah, we do this (almost) everywhere we use the `visuallyhidden` stuff.  Totally could be a mixin.

pcraig3:
Gonna merge this because it's been approved.

pcraig3:
Also it needs to be ready super soon.

PR #318, opened 2017-03-06
Noscript for multiquestion deps

samuelhwilliams:
* Add noscript tags to 'hidden' questions, which are currently used solely for question-with-dependencies followups in G9 services.
* Bump version to 22.1.0


TheDoubleK:
Should we add these for all the question types?  I know these might be the ones that are followups right now, but it seems possible that in future that other question types could be followups too.

samuelhwilliams:
I think the ones we've added it to are the main types, but have added it to pricing, upload and checkbox-tree.

We probably want to implement this differently in an ideal world, but I suppose consistency across all questions is still desirable in the short-term.

TheDoubleK:
:+1:

PR #320, opened 2017-03-06
makes question heading bold, improves legibility of follow-up questions

ralph-hawkins:
changes follow-up question headings from:

screen shot 2017-03-06 at 17 37 05

to: 

screen shot 2017-03-06 at 17 36 46


pcraig3:
Good idea.

I've done this as well, and added some examples for it here: https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/321

PR #322, opened 2017-03-14
Fix bug where two elements controlling a single target might unintentionally hide the single target

pcraig3:
If two controls are controlling one target, the last action wins.
It's possible to hide our target unintentionally.

For example, if we have two elements,
"A" and "B", both pointing at "C",
then we can hide "C" with "B",
even if "A" still says it should be revealed.

Like
A -> (show) -> C
B -> (hide) -> C

Then "C" will be hidden, even though we don't want it to be.
We only want it to be hidden if both "A" and "B" want to hide it.

This behaviour has now been corrected.

***

## gifs

showhides


samuelhwilliams:
Testing functionality, looks good.

multi followups

Issue #311, opened 2017-02-27
Links on the forms page 404


quis:
http://alphagov.github.io/digitalmarketplace-frontend-toolkit/forms/

links to:
- http://alphagov.github.io/forms/selection-buttons.html
- http://alphagov.github.io/forms/list-entry.html
- etc…

pcraig3:
Also there's no way to get back to the index page from the forms page.

pcraig3:
Hey @quis.

https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/323

carolinegreen:
Think this is fixed, the links now work so closing this.

PR #323, opened 2017-03-15
Fix broken links in generated pages

pcraig3:
[It was pointed out recently](https://github.com/alphagov/digitalmarketplace-frontend-toolkit/issues/311) that the links on the forms page in the github pages [`/forms/index.html`](http://alphagov.github.io/digitalmarketplace-frontend-toolkit/forms/index.html) page were all wrong.  Same deal with the "Home" link at the top of the page.

They worked locally, but not once they were generated online.

Now they work everywhere.

🎉


carolinegreen:
👍 Looks legit.

PR #324, opened 2017-03-22
Multiquestion dependency layout

galund:
Layout changes to make the grey bars line up into a single grey bar, creating an illusion of a "group of questions", where in fact we don't have markup that groups the related questions together :-(


TheDoubleK:
Before and after screenshots are traditional for front-end change PRs :)

toolkit/scss/forms/_combinations.scss
```diff
@@ -24,6 +28,7 @@ $border-width: $gutter / 6;
   }
 }
 
-fieldset.related-information {
-  padding-top: 0;
-}
+.related-information + .related-information {
+  padding-top: $gutter-one-third;
+  margin-top: 0;
+}
```
TheDoubleK: Missing line terminator here.

pcraig3:
Hey @galund, I've found a new weird problem.

Our final example on the Combinations page doesn't quite have the borders line up still, although now it looks really squished together.
I think it's probably the list-entry element, although I haven't really looked into it yet.

### -> old <-
screen shot 2017-03-23 at 16 09 59

### -> new <-
screen shot 2017-03-23 at 16 10 18

galund:
@TheDoubleK yep, where we previously used the fieldset for styling and interaction (by adding classes, id, etc), we now just wrap it in a div and use that instead. The fieldset becomes a purely semantic thing, so that we can mostly ignore its weird / uncontrollable layout.

Screenshots attached (tho I tend to think that checking the actual resulting layout is a QA / Product thing, and what we should concentrate on here is the code).

Before:
screen shot 2017-03-24 at 10 55 30

After:
screen shot 2017-03-24 at 10 54 02

galund:
@pcraig3 yep, fixed the list entry now. AFAICS list entry and selection buttons are the only two templates in `toolkit/templates/` that use `frameset` in that way, except:

 - checkbox-tree - we dealt with that separately
 - pricing - doesn't seem to have the option on being a 'related' / 'combination' question (is that on purpose @allait?)

Feels like there's some common stuff between the various form templates that could be factored out - will leave that as a Core Team thing, or Catalogues can come back next time we have a new feature that touches this code :)

pcraig3:
Hey @galund, I think this looks good now after a version bump.

Pricing is kind of a weird pattern anyway.

PR #325, opened 2017-04-04
141883849-Create-date-input-template

benvand:
## Create template

* Add SCSS classes to allow for sizing and padding small text boxes.
* Create template using common elements from other templates.
* `summary-table` and `summary-content` macros

As per:
https://www.gov.uk/service-manual/design/dates#asking-for-memorable-dates

https://trello.com/c/I7bAJzq9/34-create-date-input-template
### Before:

old-date-field

### After:
new-date-field

### Summary table:
stable

### Pages Builder documentation:
screen shot 2017-04-10 at 13 18 19


toolkit/templates/forms/date.html
```diff
@@ -0,0 +1,86 @@
...
+    <fieldset class="text-box-inline-container" id="{{ name }}" >
+
+      {% for display_name, field_name, field_length, field_class in
+        [
+          ("Day", name + "-day", 2, "text-box-tiny"),
+          ("Month", name + "-month", 2, "text-box-tiny"),
+          ("Year", name + "-year", 4, "text-box-small")
+        ]
+      %}
+        <div class="text-box-inline-rp">
```
carolinegreen: What does rp stand for?

carolinegreen:
Nice you are using the frontend toolkit.  It would be good if you added screenshots to see what this new template looks like.  Unfortunately, I don't have access to that trello board yet, so my context is limited, but will look again when I do.

benvand:
'right-padded'. Could extrapolate that into its own class or rename to
include it if you'd prefer?

galund:
It's much better if class names can correspond the purpose of their content, rather than the way that you're currently planning to lay them out?

benvand:
@galund I'm concerned that makes for a lot of repetition. I was looking to reuse some of the stuff from the pricing field for this but it wouldn't have made sense given that this is the date field.

toolkit/templates/forms/date.html
```diff
@@ -0,0 +1,86 @@
...
+          <label for="input-{{ field_name }}">{{ display_name }}</label>
+          <input
+            id="input-{{ field_name }}"
+            name="{{ field_name }}"
+            type="text"
```
idavidmcdonald: Should this be `type="number"`?

TheDoubleK:
_mumbles_ ...something something semantic class names something...

toolkit/templates/forms/date.html
carolinegreen: No, this is right.  Number isn't supported by ie9.

carolinegreen:
The screenshots make it a lot clearer, thanks.  For the summary macro table, why is this being touched, where is this being used, and do you have an example of this too?

@galund @TheDoubleK Yes it's better if it's semantic :sparkles:  @benvand You could try describing what it's for (dates) and why there are two versions of the textboxes.  I take the padding is for the split date?

This is also more of a general thing, but I'm wondering if we should be linking to trello cards.  They can be very transient compared to git pull requests and if someone is looking at this say, a year in the future trying to work out the reasoning behind the code then I think linking to a card can miss a lot of the context.

benvand:
NP, added summary screenshot.

@everyone
Strictly speaking the scss is for text boxes. It's in the `_textboxes` file.
Don't get why we'd limit this to only being used for dates? That would make it single use.

Re Trello:
That's more of a me thing than anything else. The _only_ thing we can really rely on persisting is commit comments.
I find it useful to link to the ticket for the current context. Who's doing it? Who's discussing it? What state is it in?
Rather too much info than too little.

toolkit/templates/forms/date.html
idavidmcdonald: Serious front end knowledge there

idavidmcdonald:
Should we also add the date field to the docs with some examples? http://alphagov.github.io/digitalmarketplace-frontend-toolkit/

carolinegreen:
@benvand What is the reason for it being inline and having padding?  Is it to do with a design pattern for dates or is this pattern currently being used somewhere else?  

From what I can see it looks like these textboxes are being used for dates, and if these textboxes are used in another context then we can look at the commonality between them and come up with another name more suitable to that situation.  The dates name is also just a suggestion; some kind of semantic name would be better.

For trello, you can link to git pull requests on trello cards, which makes sense for team context.  On git, people who don't have access to that trello board can't see what's on it, which makes the request seem like it has extra hidden information and not very open.

@idavidmcdonald's suggestion is a good one too.

benvand:
@carolinegreen 
The field was designed as per the service manual.
Can I get some suggestions for the class names?
https://www.gov.uk/service-manual/design/dates#asking-for-memorable-dates

benvand:
@idavidmcdonald 
See PR description for Pages Builder docs

toolkit/scss/forms/_dates.scss
```diff
@@ -0,0 +1,12 @@
...
+  padding-right: 1em;
+}
+
+.date-input-day-month {
+  width: 2.75em;
+}
+
+.date-input-year {
+  width: 4.25em;
+}
```
carolinegreen: nit - newline

carolinegreen:
Apart from one tiny nit looks good to go 👍

pages_builder/assets/scss/forms/index.scss
```diff
@@ -5,6 +5,7 @@ $path : '../../images/';
 @import "shared_placeholders/_placeholders.scss";
 
 @import "forms/_textboxes.scss";
+@import "forms/_dates.scss";
```
carolinegreen: Should this be being pulled in in two different places?  Some of the other stuff that's duplicated there already looks quite old.

toolkit/scss/forms/_dates.scss
benvand: Fixed

pages_builder/assets/scss/forms/index.scss
benvand: Fixed

idavidmcdonald:
Cool, this looks good and I see no outstanding comments from @carolinegreen so I'll give a thumbs 👍 

Note, it needs a version bump before it can be merged though

PR #327, opened 2017-04-13
Fixes validation bug

MuriloDalRi:
[Story](https://www.pivotaltracker.com/story/show/143115739)

Checks if the link in the validation masthead is an anchor link before running the event listener function.

This allows external links such as 'mailto' ones to work as expected.


risicle:
Neat.

Wonder if we should add a regression test for this... I haven't ever looked at the way the tests work in this repo.

carolinegreen:
I think historically we've been pretty bad for writing tests in the frontend toolkit on digital marketplace.  @pcraig3 probably has more context on the current state of the tests currently.  You could write a jasmine unit test to check that you are jumping from the error summary to the field, which would probably be good practice.  

We have some of those for the show hide javascript for digitalmarketplace-frontend-toolkit/toolkit/javascripts/show-hide-content.js in spec/javascripts/unit/show-hide-content.spec.js

pcraig3:
Yeah, seems like it would be easy to write a simple test that checks the focus moves to the element you want it to.  

If you have a look in [the spec/javascripts/unit](https://github.com/alphagov/digitalmarketplace-frontend-toolkit/tree/master/spec/javascripts/unit) folder that @carolinegreen mentioned, you can try to emulate what they're doing.  Alternatively, if that seems a bit daunting, one of us can help you get set up when we're back in the office.

Hopefully this isn't an emergency that needs to be pushed out today or anything. 

Final point is you'll want to bump the version so that the frontend apps can bring this in.

TheDoubleK:
It's not an "emergency" as such, but it is a bug in production so would be nice to get it out at some point...

pcraig3:
We're on it, Kev. Hold your horses.

TheDoubleK:
:hand: :horse: :horse: :hand:

pcraig3:
👍  Do it.

PR #328, opened 2017-04-19
Set zero padding on margin bottom only

carolinegreen:
Set zero padding on margin bottom only as setting it on every side overrides required padding on other sides.

Before:
screen shot 2017-04-19 at 16 48 11

After:
screen shot 2017-04-19 at 16 47 38


pcraig3:
Easy 👍

PR #329, opened 2017-04-20
Remove padding on bottom radio combination

carolinegreen:
Remove padding on bottom radio combination as this extends the grey connection border past the radios.  Also the pseudo selector previously used for this isn't compatible with IE8 which we're still supporting.

Before:

screen shot 2017-04-20 at 13 26 12 1

After:

screen shot 2017-04-20 at 13 24 49


pcraig3:
So [I've left a comment](https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/330#discussion_r112718929) on #330 about this particular change, recommending against doing it.

This CSS is trying to put more space between chained follow-up questions so that they're more clearly differentiated from each another when there's a lot of them.
In your screenshot, you show the un-expanded version of the follow-up question.  When you expand it so you can see both follow-up questions, I think the [before](https://cloud.githubusercontent.com/assets/2454380/25285330/bd771946-26b1-11e7-9531-eb0732eb57bc.png) looks better than the [after](https://cloud.githubusercontent.com/assets/2454380/25285346/c9c7837a-26b1-11e7-8db6-fb20e6678dfb.png).

Not against removing this stuff for something more robust -- I didn't know that IE8 wouldn't respect these rules -- but I don't think we should remove it without an alternative.

carolinegreen:
I'm going to close this one as I think all these problems are too connected to fix this with just looking at these particular lines.  I think replacing this code will fix a lot of the other visual bug problems too.

PR #330, opened 2017-04-21
Add last of type class

carolinegreen:
The :last-of-type pseudo selector isn't supported by ie8 so add last-of-type class in macro.  Last-of-type css class is being added within checkbox scss for now as this seems specific to this case rather than a general reset.

The new margin now means that when the focus is on a checkbox, the yellow focus border extends beyond the grey connection border.  However when there is no focus the checkboxes are now aligned with grey border, matching service manual patterns.

Before:
screen shot 2017-04-21 at 16 11 25

After:
screen shot 2017-04-21 at 15 31 47


toolkit/templates/forms/_checkbox-tree-selection-buttons.html
```diff
@@ -25,12 +25,18 @@ <h3 class="categories-heading">{{ parent_category.label }}</h3>
             </legend>
 
             {% for option in parent_category.options %}
+              {% if loop.last %}
```
pcraig3: You could probably pass `loop.last` directly in as a variable, or even `set is_last = loop.last` or something to save a bit of space.

carolinegreen: Yeah, that was overly verbose.

pcraig3:
Still gets cut a bit short on mobile.  

image

😦

carolinegreen:
Yeah was fixing that as you wrote that, doesn't anymore!

toolkit/scss/forms/_combinations.scss
```diff
@@ -13,10 +13,6 @@ $border-width: $gutter / 6;
   padding-top: $gutter-two-thirds;
   margin: -$gutter 0 0 0;
 
-  &:not(:last-of-type) {
-    padding-bottom: $gutter;
-  }
-
```
pcraig3: I'm not sure about cutting out these rules.  There are two things it seems to make worse right now.

| Before | After |
|--------|-------|
| screen shot 2017-04-21 at 16 38 59 | screen shot 2017-04-21 at 16 40 41  |
| image | screen shot 2017-04-21 at 16 42 08 |

In the first case, we lose spacing between the first and second questions and we break the continuous grey border.  
In the second case, the follow-up questions (once they're both opened) look a bit too squashed together.

carolinegreen: That's pretty weird that the grey border isn't continuous and suggests a markup problem.  Also this isn't compatible with ie8 which we still support, so we should be finding an alternative anyway.

carolinegreen:
I think these are both fixed now and ready to review.

risicle:
OT but what do we think of polyfills for this kind of thing? Suppose it only works with js enabled...

carolinegreen:
@risicle I think you answered your own question there :P

pcraig3:
Hey, I only properly reviewed this at end of day today. 
I have a few comments and I'm not sure if this is the best place for them but it seemed like it would be fine.

carolinegreen:
Probably a bit late :s  Want to talk about it tomorrow?

pcraig3:
These are just changes that I noticed doing a manual visual diff, not looked at any of the CSS changes in particular.  Wraith just turned the whole screen red for me (because all the spacing is changed), so it wasn't really that useful.

## i. There's a new gap in the border when we have textareas with wordcounts

| Before | After |
|--------|-------|
| screen shot 2017-05-04 at 19 07 48 |  screen shot 2017-05-04 at 19 08 18 |

## ii. Questions are generally more squished together

This is probably down to a matter of preference, but I think it was better before having them more clearly separated.

| Before | After |
|--------|-------|
| screen shot 2017-05-04 at 19 09 08 | screen shot 2017-05-04 at 19 09 26 |

## iii. More space between the question and hint_underneath fields

The `hint_underneath` is further from the question it's related to and closer to the next one. 

| Before | After |
|--------|-------|
| screen shot 2017-05-04 at 19 15 14   | screen shot 2017-05-04 at 19 15 32  |

## iv. There's a lot of extra spacing when radio buttons/checkboxes have extended descriptions

To be clear, this isn't a new problem.  It's been like this for a while.  
The difference now is that the fields are more compressed though.

| After |
|-------|
| screen shot 2017-05-04 at 19 11 07  The different coloured lines are the distances between the bottom of the field above and the top of the bolded text about the next question.|

## v. Border extends further under vertical radios/checkboxes

We already knew about this.

| Before | After |
|--------|-------|
| screen shot 2017-05-04 at 19 16 26 | screen shot 2017-05-04 at 19 16 42 |

***

I think probably i, ii, and iii are specific to this pull requests.  It also doesn't seem like any of them are showstopping bugs.

carolinegreen:
I think i. is a bug worth fixing and putting in the backlog.  The other ones seems pretty negligible.  v isn't nice, but there are other problems with the vertical lines that are probably a separate thing.

PR #331, opened 2017-05-02
Add IE8 compatible CSS for breadcrumbs

carolinegreen:
Use :first-child instead of :last-child pseudo class which is supported by IE8.  No visual changes.


toolkit/scss/_breadcrumb.scss
```diff
@@ -28,17 +28,17 @@
...
-        margin-left: 0;
         margin-right: 0.5em;
-        padding-right: $gutter-half;
+        padding-left: $gutter-half;
 
-        &:last-child {
+        &:first-child {
           background-image: none;
-          margin-right: 0;
+          margin-right: 0.5em;
```
risicle: Is this line redundant? (already specified above)

carolinegreen: Yes, good spot.

PR #332, opened 2017-05-05
[WIP] New example page with lots of child questions

pcraig3:
One of our problems around borders for child questions has been that we just don't have a view of very many of them.

There are far more combinations of child/dependent questions in the submissions process than we actually have in the frontend toolkit in the 'combinations' page, yet it's the combinations page that we're normally checking when we're doing CSS changes.

Essentially, we're visually testing a very small subset of the total universe of possibilities and so we're going to get things slip through.

This code creates a new page on which we can very rapidly create loads of child fields one after another.

It's pretty prototype-y code (and the new page isn't included in any of the menus), but it's helpful for seeing the effect of general CSS changes to child questions.

***

This is what it looks like if you boot up the toolkit and run: http://localhost:8000/forms/all.html

screen shot 2017-05-05 at 11 51 28-fullpage

### DIFF
--- a/pages_builder/pages/forms/all.yml
+++ b/pages_builder/pages/forms/all.yml
@@ -1,0 +1,98 @@
+pageTitle: All fields
+assetPath: ../govuk_template/assets/
+grid: column-two-thirds
+bodyEnd: >
+  <script type="text/javascript" src="../public/javascripts/govuk_frontend_toolkit/vendor/polyfills/bind.js"></script>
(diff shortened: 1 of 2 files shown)


TheDoubleK:
I reckon it would be good to link this from a menu otherwise people will forget about it/never know it exists.

pcraig3:
I think my current story about [bringing in the macros](https://trello.com/c/LVdwGWJN/60-unify-toolkit-forms-wrappers) will have some implications for this story.

PR #333, opened 2017-05-05
Fix focus bug on list entry buttons (in FF)

pcraig3:
We've had a tiny tiny bug open for a while about focus buttons for list entry buttons in firefox specifically. @tombye mentioned a while back that there's a one line fix for it.

Boom.

***

## before

focus

## after

focus_fix


carolinegreen:
Nice! 👍

PR #334, opened 2017-05-05
Bugfix for IE8 show/hide boxes

pcraig3:
Child/dependent fields are hidden on pageload but then the JavaScript that reveals them is broken in IE8.  This means that they remain completely hidden.

### 😱

show_hide

Created a bug for this yesterday and had it sitting in the backlog, but earlier today @idavidmcdonald messaged me about a supplier who was having issues applying to a brief because they couldn't.  He sent along the following screenshot, which shows that the supplier has clicked "Yes", but the follow-up field hasn't been revealed: 

pasted image at 2017_05_05 01_50 pm

Used http://jscc.info/ to lint the file and then swapped out the couple methods that IE8 doesn't support with some jQuery helper functions. 

Tested it in IE8 on browserstack and it works now.

ie8


VERSION.txt
```diff
@@ -1 +1,2 @@
-22.2.4
+22.2.5
+
```
idavidmcdonald: What's this extra line doing in here ;)

pcraig3: Good spot, David!

idavidmcdonald:
👍 seems reasonable but there is an extra whitespace line in the version file which we should get rid of first

PR #335, opened 2017-05-05
Add ie8 compatible CSS for dmspeak

carolinegreen:
First child is supported for IE8 so reverse padding.  No visual changes.


pcraig3:
Does something funny to list items in our longer `question_advice` fields.

| Before | After |
|--------|-------|
| screen shot 2017-05-05 at 16 47 18 | screen shot 2017-05-05 at 16 47 45  |

carolinegreen:
Must be interacting with other CSS, going to close this so I can look at it.

pcraig3:
I've just tested it again and I'm definitely seeing this in Chrome, Firefox and Safari. Maybe it's a caching thing?

carolinegreen:
Yeah looks like it. I thought I had cache busted, must not have.

carolinegreen:
If you remove the negative margins, which you don't usually want, it solves that problem:

screen shot 2017-05-08 at 15 31 35

pcraig3:
Yup, looks the same as before now.  Probably better this way, actually. 👍

PR #336, opened 2017-05-05
Reverse margins for questions

carolinegreen:
Reverse margins for questions to use margin top and first child instead of margin bottom and last child as this is compatible with ie8.  No visual changes.


pcraig3:
👍

PR #337, opened 2017-05-08
Remove ie8 incompatible CSS for instruction list

carolinegreen:
Replace  pseudo selectors with question class instead, as this is what is causing the excess padding that needs to be overidden.  No visual changes.


carolinegreen:
The question class is applying the excess margin originally overridden by using the pseudo selectors.  This excess margin applied by the question class isn't required in instruction lists, so it overrides the larger margin applied by the question class for more general examples to a smaller margin for the instruction list.

pcraig3:
Cool, looks good I think.

pcraig3:
👍

PR #338, opened 2017-05-09
Standardise import paths between the apps and the frontend toolkit

pcraig3:
TL;DR

We can now do this 
```
{% include 'toolkit/forms/_selection-button.html' %}		
```

instead of this:
```
{% if PAGES_BUILDER %}		
  {% include 'forms/_selection-button.html' %}		
{% else %}		
  {% include 'toolkit/forms/_selection-button.html' %}		
{% endif %}
```

***

When linking to jinja2 partial templates, you have to give a local path, which is appended to a template root path.
Our problem is that out template paths differ depending on whether we're building pages in the toolkit itself or whether we're building pages in our apps.

As an example, here is the path to `_selection-buttons.html`:

- buyer frontend: `app/templates/toolkit/forms/_selection-buttons.html`
- frontend-toolkit: `toolkit/templates/forms/_selection-buttons.html'

Since the `/toolkit` and `/templates` directories are reversed across the repositories, it makes it hard for us to come up with a common root path + path suffix.

For example if we want to import `toolkit/forms/_selection-buttons.html`, that's fine in the buyer app, but it doesn't work in the frontend toolkit because the filepath we want there is actually `templates/toolkit/forms/..`.

This commit doesn't give us general behaviour, but it does give us a consistent way of importing things inside the toolkit itself and the frontend apps.

For any include paths, the 'toolkit/' prefix will be removed.
So {% include 'toolkit/forms/textbox.html' %} will be understood
as {% include 'forms/textbox.html' %}, which fixes our specific
problem.

We could generalise this behaviour further, but at the moment we don't need to.

Found this helpful: http://stackoverflow.com/a/3655911


pcraig3:
[Take out the weird `if` statements in jinja templates](https://trello.com/c/LbtVkIGf/80-take-out-the-weird-if-statements-in-jinja-templates)

toolkit/templates/forms/_checkbox-tree-selection-buttons.html
```diff
@@ -34,14 +34,7 @@ <h3 class="categories-heading">{{ parent_category.label }}</h3>
...
-
-                {# TODO: PLS FIX #}
-                {% if PAGES_BUILDER %}
-                  {% include 'forms/_selection-button.html' %}
-                {% else %}
-                  {% include 'toolkit/forms/_selection-button.html' %}
-                {% endif %}
-
+                {% include 'toolkit/forms/_selection-button.html' %}
               {% endwith %}
```
carolinegreen: Do we want toolkit in front of everything?  In your main pr you suggest that we're going with the other option which would be `{% include 'forms/_selection-button.html' %}` in this case.

galund:
There's a possible answer on SO which is more general in terms of allowing all relative paths to be resolved - http://stackoverflow.com/a/3655911/297790

toolkit/templates/forms/_checkbox-tree-selection-buttons.html
pcraig3: So this pull request would silently remove the 'toolkit/' part of include filepaths, but only when building pages for the toolkit itself.

ie, if you call this
```
 {% include 'toolkit/forms/_selection-button.html' %}
```
it will actually look for the path at `forms/_selection-button.html` when it's building the pages. 

It's a bit gross, but it allows us to:
- use the same path across both the frontend toolkit and the frontend apps
- not need any code changes in the frontend apps themselves

The other option would be to use relative paths, which is what George is suggesting below.

pcraig3:
Yeah, I linked to the wrong answer in my pull request above, @galund, but that's the one I meant to link to. 

Using relative paths would be technically cleaner, but it seems like the reason that relative paths aren't allowed by default are because of the risk of exploiting this behaviour to look through your file system.  I haven't found anything in the jinja codebase itself, but there's [a comment in the ansible repository which talks about exactly this problem](https://github.com/ansible/ansible/issues/14992#issuecomment-274563707). 

I figured that this gives us the behaviour we wanted without exposing us to * risks *, but if we'd rather do the relative path thing, then I'm happy to make the change.

toolkit/templates/forms/_checkbox-tree-selection-buttons.html
pcraig3: My other pull request builds some internal pages that are useful for testing layout changes, so the include paths in there don't need to be usable in both contexts.  I put up the pull request to show what the new code is doing, but I think at this point it's probably better to close it until the macros refactoring is finished.

galund:
So I guess there's a more complicated solution, where relative paths are allowed, but they have to resolve to something inside the original base path. It's probably a one-line check, and for my money that'd be real nice to have - but definitely this PR is much improved on what we have now.

toolkit/templates/forms/_checkbox-tree-selection-buttons.html
carolinegreen: So why do you have this at the top of the pull request?

screen shot 2017-05-12 at 10 02 00

pcraig3: Ah right, that's a mistake.

carolinegreen:
I think we should merge this, especially if relative paths are potentially risky.  It is much nicer than what we have currently! 👍

PR #339, opened 2017-05-09
Remove padding last child for hints

carolinegreen:
This is redundant as this extends from markdown where the padding has already been added to the top rather than the bottom, so there is no bottom padding already.  At some point we should start removing the extends for various reasons, but this is outside of the scope of this pull request.  No visual changes.


pcraig3:
👍

PR #340, opened 2017-05-09
Change last child to first child(ie8)

carolinegreen:
Change :last-child to :first-child and padding from left to right, so that the css is compatible with IE8.  No visual changes.


pcraig3:
👍

PR #341, opened 2017-05-10
Use CSS selectors compatible with IE8

carolinegreen:
Use CSS selectors compatible with IE8 for checkbox trees.  Add padding to top of adjacent details to details with the attribute of open instead of using pseudo classes.  Remove padding at bottom of paragraph class is it isn't having a huge effect and doesn't line up perfectly as intended.  If we want this to line up perfectly we should look at this later and try to apply this style another way.


pcraig3:
This is definitely cleaner CSS 

👍

PR #342, opened 2017-05-12
Reliably kill multithreaded server

pcraig3:
When I was trying to get Wraith to work, I updated our little Python server to support multiple threads.
This meant that Wraith could open up a bunch of pages simultaneously without the thing crashing out.

Unfortunately, this made killing the server with "Ctrl+C" extremely unreliable.  Lots of times it would just hang until the terminal pane was closed, which was immensely frustrating.

I think this commit fixes it.
Helpful source: http://kmkeen.com/socketserver/2009-02-07-07-52-15.html


pcraig3:
Always on the cutting edge here at GDS, Kev.

PR #343, opened 2017-05-12
Add secondary search button to keyword search box

TheDoubleK:
For this story: https://trello.com/c/dtI7Ozur/414-keyword-search-bar-location

The buyer frontend isn't currently using the keyword search pattern from the toolkit. As part of the story to move and update the search box it seemed worth making the style we want in the toolkit and actually using it.

This also brings our search box into line with the current GOV.UK "secondary search box" style you can see here: https://paper.dropbox.com/doc/Search-boxes-WVuvy5JmNJLK9w4Tq4QHC

It looks like @pcraig3 did all the work here, but rest assured I did a bunch of very important preliminary work too :stuck_out_tongue: 

## Existing toolkit
screen shot 2017-05-12 at 14 16 48

## How it currently looks in the buyer frontend
screen shot 2017-05-12 at 14 17 03

## How it looks with the changes here
screen shot 2017-05-12 at 14 17 28


toolkit/scss/forms/_keyword-search.scss
```diff
@@ -2,10 +2,83 @@
...
+    background: #fff;
+
+    &.focus,
+    &:focus {
+      outline-offset: -2px;
+    }
+
+    .js-enabled & {
+      @include calc(width, "100% - 36px");
+    }
```
pcraig3: Actually, in light of your earlier question about what `.js-enabled` is actually doing for us, I don't think we need this. 

Presumably, even without javascript running, we still want the `calc` rule to work on browsers that support it.  I think you're probably fine to get rid of this `.js-enabled &` rule and move the calc declaration just after the `width` rule above.

TheDoubleK:
Yep - removed it and it still works OK.

I'm happy to merge it if you're happy to thumbs it up - QA'd to the max by ourselves.

toolkit/templates/forms/keyword-search.html
```diff
@@ -1,8 +1,11 @@
-<div class="govuk-option-select">
+<div class="govuk-option-select filter-field-text">
   <div class="container-head">
     <label class="option-select-label" for="{{ id }}">
       {{ label }}
     </label>
   </div>
-  <input class="filter-field-text" type="text" name="{{ name }}" id="{{ id }}" value="{{ value }}" maxlength="{% if maxLength %}{{ maxLength }}{% else %}200{% endif %}">
+  <div class="container-search">
+    <input type="search" name="{{ name }}" id="{{ id }}" type="{{ label }}" value="{{ value }}" maxlength="{% if maxLength %}{{ maxLength }}{% else %}200{% endif %}">
```
pcraig3: Oh no! Final point! (I thought I removed this but I guess it didn't stick!)

There are two `type` attributes.  We want to keep the one that says type="search" and not the other one.

TheDoubleK: :eyes:

carolinegreen:
I think maybe a third person should review it if it's not got to be out before five just because you have both worked closely on it and it would be a fresh pair of eyes.

TheDoubleK:
So sensible 😛

samuelhwilliams:
I've added an extra commit that tweaks the border around the keyword search to wrap around the search icon as well, which @pcraig3 and I agree looks a little nicer.

## Old
screen shot 2017-05-15 at 12 24 58

## New
screen shot 2017-05-15 at 12 24 19

toolkit/scss/forms/_keyword-search.scss
```diff
@@ -2,10 +2,80 @@
...
+  input[type="search"] {
+    @include core-16;
+    @include box-sizing(border-box);
+    position: relative;
+    padding: 6px;
+    width: 86%;
+    @include calc(width, "100% - 36px");
+    border: 1px solid $border-colour;
+    border-right: none;
+    background: #fff;
```
carolinegreen: I'm not sure if the background color is adding anything in this case.

toolkit/scss/forms/_keyword-search.scss
```diff
@@ -2,10 +2,80 @@
...
+    background-repeat: no-repeat;
+    background-position: 0 50%;
+
+    @include device-pixel-ratio() {
+      background-size: 52.5px 35px;
+      background-position: 100% 50%;
+    }
+
+    &:hover {
+      cursor: pointer;
```
carolinegreen: Do we want this to be a pointer?

pcraig3: [The magnifying glass on the homepage](https://www.gov.uk/) uses a pointer.
[The magnifying glass on the general search page](https://www.gov.uk/search?q=sponge+avengers) uses a pointer. 
[The magnifying glass on the "browse" search page](https://www.gov.uk/browse/driving) doesn't use a pointer.  

Decided to go with 2/3.

carolinegreen: Yeah actually got mixed up forgot pointer wasn't an arrow, pointer is better.

toolkit/scss/forms/_keyword-search.scss
TheDoubleK: No I'm pretty sure it's superfluous - have removed it now.

PR #344, opened 2017-05-15
Allow users of our summary table macros to control bold

galund:
Requirement here was for the questions with dependencies to be bold, on the buyer side (service page).
 - for that, we need to be able to get the 'bold' param passed all the
   way down to the bottom-level field macro.
 - this is pretty ugh and demonstrates a flaw in the way we are using
   macros, arguably.

https://trello.com/c/9mmLtRCC/352-questions-with-dependencies


CHANGELOG.md
```diff
@@ -2,11 +2,20 @@
 
 Records breaking changes from major version bumps
 
+## 22.4.0
```
samuelhwilliams: Only required to update the changelog for breaking changes... But I'm not especially opposed to more documentation. :)

galund: oh, good point :)

pcraig3:
Probably out of scope for this story, but since we have lots of rarely-used optional fields in some kind of base macro, we could just pass `kwargs` all the way through and then only check in the final field for the actual keys we're looking for.

For example 

```jinja
{# simplified field macro #}
{# note that you can't use `kwargs` in the parameter definition, but you can refer to it in the function #}
{% macro field(rowspan="") -%}
  {% set data = caller() %}
  <td {% if rowspan %}rowspan={{ rowspan }}{% endif %} class"...">
    {% if data %}
      {% if kwargs.bold %}<strong>{% endif %}<span>{{ data }}</span>{% if kwargs.bold %}</strong>{% endif %}
    {% endif %}
  </td>
{%- endmacro %}

{# simplified text macro #}
{% macro text(text='', assurance=None, rowspan='') -%}
  {% call field(rowspan=rowspan, **kwargs) -%}
    {% if text -%}
      {% if kwargs.large %} This is large {% endif %}
      {{ content.text_content(text, assurance) }}
    {%- endif %}
  {%- endcall %}
{%- endmacro %}

{% macro textbox_large(value='', assurance=None) -%}
  {{ text(value, assurance, large=True, **kwargs) }}
{%- endmacro %}
```

This would let you do something like this:
```jinja
{{ text("text that I want to be bold", bold=True) }}
{{ textbox_large("large textbox that I want to be bold", bold=True) }}
```

The way that `kwargs` is implemented in jinja is, arguably, a [distasteful design choice](http://stackoverflow.com/questions/13944751/args-kwargs-in-jinja2-macros#comment19229904_13944920), but we could implement it once in all the intermediary macros and it would would save us from having to change tons of parameters the way you're doing here.

galund:
Yeah, I'm not against that completely (agree it's ugly, but everything about macros seems to involve ugly).

I think I'd argue the whole thing needs re-thinking, and if we're doing that, then we might not be using macros at all, in which case it's moot.

carolinegreen:
Using kwargs does seem more elegant, although you still have the problem of things being hidden, and currently having the options on the top macro let's you see immediately what is possible to put into that macro before you write one.  If we were to do this we would have to have some other way of making it clear what options are available on the uppermost macro.

@george Are you changing all the macros to make things more flexible or is the bold in all these fields?  I think one of the reasons in the past that we didn’t add bold to everything was that bold things would be part of a design pattern so having different things emphasised would be a different design pattern with different meaning.  This is to avoid us having inconsistencies across the apps with different designs for similar things.  Do you have an example of where this is being used?  Or are you waiting for the request to be approved to pull it in?  If it’s quite different from the other tables could you add an example to the toolkit?

galund:
So the purpose of this is to try to show some structure, so that users can get information quickly by scanning down the page. A question with followups is 'more important' than the questions which follow, so the question _and the response to that question_ were going to be bold.

As I was passing in bold=False for the questions don't need to be bold, passing that param needed to be possible across all the question types.

The code (that isn't done with yet, as it doesn't actually quite do the right thing) is here: https://github.com/alphagov/digitalmarketplace-buyer-frontend/compare/questions-with-dependencies-service-page

But you might find that @JeremyHHY has a mockup that makes it more clear.

There's no particular reason why we wouldn't want the service summary for buyers to look the same as the service summary for suppliers, but right now they have different templates (albeit the share most of the markup via the macros).

carolinegreen:
@galund yeah that makes sense and is a nice explanation.  If you don’t want to put this upstream as a pattern example on the summary tables page in the toolkit at this stage could you pull in a branch of the toolkit in the meantime and add a screenshot to the pull request in the buyer frontend to show how these changes to the fields in the toolkit are being used.  I realise this is slightly more work, but it means that we have a visual record of that design pattern, which gives more context, is less likely to be lost than a mockup (e.g. if we’re revisiting this years later), and is also actually in code in the buyer app.

galund:
Closing pending further discussion about the look and feel of the service summary page.

PR #345, opened 2017-05-16
Remove IE8 incompatible CSS

carolinegreen:
Reverse margins and remove :last pseudo selectors.  Also refactor some of the CSS to make it simpler.  We should change the checkbox__tree classnames as this doesn't reflect the markup used in the table and is confusing, however this is outside the scope of the current pull request.  No visual changes.


TheDoubleK:
Looks good as far as I can tell! :+1:

pcraig3:
Yeah, 👍  from me too if you were waiting for it. Looks good!

carolinegreen:
Was doing interviews!  Thanks both!

PR #347, opened 2017-05-17
Remove `visuallyhidden`. Use `visually-hidden` instead.

pcraig3:
This change is super trivial.

Technically, both `visuallyhidden` and `visually-hidden` will work
as classnames to hide things in the visual interface (both are
included in govuk_template), but here in _our_ frontend toolkit,
we should be consistent.

`visually-hidden` is more readable and closer in convention to
our other classnames in, so I've removed the other ones.

I've also removed a `visually-hidden` CSS rule that was just a
straight copy from something we already get from govuk_template.

No visual changes.

### DIFF
--- a/toolkit/javascripts/list-entry.js
+++ b/toolkit/javascripts/list-entry.js
@@ -32,7 +32,7 @@
   ListEntry.prototype.entryTemplate = Hogan.compile(
     '<div class="list-entry">' +
       '<label for="{{{id}}}" class="text-box-number-label">' +
-        '<span class="visuallyhidden">{{listItemName}} number </span>{{number}}.' +
+        '<span class="visually-hidden">{{listItemName}} number </span>{{number}}.' +
       '</label>' +
       '<input' +
         ' name="{{name}}"' +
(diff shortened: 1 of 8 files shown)


carolinegreen:
Why did you choose the one without the hyphen?

pcraig3:
It's not so much the one I chose as the one we've chosen over time. [The one without the hyphen is all over our codebase](https://github.com/alphagov/digitalmarketplace-frontend-toolkit/search?utf8=%E2%9C%93&q=visuallyhidden&type=), whereas the one with the hyphen only had a couple of hits.

carolinegreen:
So really not trying to be annoying, but I find the hyphen makes it easier to read, and seems more consistent with our naming conventions for other classes.

pcraig3:
Yeah, fair enough. I guess there's not really that many of them.

carolinegreen:
I think making them consistent is a good idea though!

pcraig3:
Cool, I'll remove the `WIP` tag once I've made the change.

pcraig3:
Changed! Ready to be reviewed again.

carolinegreen:
Nice 👍

PR #349, opened 2017-05-19
Revert broken changes

TheDoubleK:
This was a bad idea: https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/348

So let's not do it.


pcraig3:
👍  once it's "in date".

PR #351, opened 2017-05-26
Unify toolkit forms wrappers

pcraig3:
The idea here is to combine the forms macros that were initially copy+pasted across the [buyer](https://github.com/alphagov/digitalmarketplace-buyer-frontend/blob/master/app/templates/macros/toolkit_forms.html), [supplier](https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/master/app/templates/macros/toolkit_forms.html), and [admin](https://github.com/alphagov/digitalmarketplace-admin-frontend/blob/master/app/templates/macros/toolkit_forms.html) apps and have been slowly going out of sync ever since.

This pull request doesn't do anything by itself.  The more interesting part of it is tearing the macros out of the frontend apps and seeing that everything still works as expected, which probably means just manually running through framework applications.  

[Here's the Trello card for this](https://trello.com/c/LVdwGWJN/60-unify-toolkit-forms-wrappers).

***

**Note** 

There seem to be a few little gotchas between the various macros.  

1. Some use extra parameters (ie, `question_number=None`) while others have been converted to using `kwargs`.  I've gone with `kwargs` in here; they _should_ be a drop in replacement for the named parameters versions, but I guess it depends on argument ordering.

2. The supplier app has a `|question_references` jinja filter, which means it will take a string like `"answer this question if you answered question [[questionId]]"` and turn that into `"answer this question if you answered question 23"`.  Since it's just doing a variable substitution, [I've done some work in the content loader](https://github.com/alphagov/digitalmarketplace-content-loader/pull/44) so that we can just use jinja rendering for that stuff and then we can strip [`question_references`](https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/master/app/main/helpers/frameworks.py#L155) out of the supplier app forever.

***

**Second note**

I've tried [removing the `toolkit_forms.html` file in the buyer app](https://github.com/alphagov/digitalmarketplace-buyer-frontend/tree/pc-make-master-macros) and that seems to have worked just fine.


pcraig3:
[Unify toolkit forms wrappers](https://trello.com/c/LVdwGWJN/60-unify-toolkit-forms-wrappers)

carolinegreen:
@pcraig3 I don't see where the assurance questions are being used in the old apps.  All I can see are the macros, and could do with examples if there aren't already.  I could do with some context on this and would ask you.

pcraig3:
💥

PR #352, opened 2017-05-26
Bugfix: date fields aren't displaying correctly in the buyer app

pcraig3:
Date fields aren't being styled correctly because the `.scss` files were re-ordered in [this commit](https://github.com/alphagov/digitalmarketplace-buyer-frontend/commit/feccdde0f32c6449e07feb7017fab5c14b4a5c76).

Making the styles more specific would make sure the styles aren't dependent on the order the files are imported.

[Trello card](https://trello.com/c/XCy5T0uU/160-date-fields-are-too-long).

## the fields are too wide 😞
image


pcraig3:
[Date fields are too long](https://trello.com/c/XCy5T0uU/160-date-fields-are-too-long)

TheDoubleK:
Sorry about that :/

But we really shouldn't rely on order of imports for anything.

This looks legit :+1:

PR #353, opened 2017-06-19
Bring in non javascript buttons and elements as a dependency

carolinegreen:
Bring in Elements SASS as a dependency rather than copying and pasting as this allows us to see more clearly what SASS we're bringing in and which SASS we're using that's custom.  It should also mean that if elements changes it is easier for us to update.

Update the markup to use the latest markup used in elements for the non javascript checkbox and radio buttons.  This also allows us to get rid of selection buttons js.  Some changes to show hide were also required as we are using a custom version of GOV.UK's show hide at the moment.

This affects quite a few of our form elements, but not option select as this would also require a redesign to use the new buttons and is outside the scope of this request.


pcraig3:
This mostly looks good, but there's been a regression with selection buttons on IE8.  

Checkboxes/radio buttons without followup questions aren't displaying at all anymore on IE8 for some reason.


| checkboxes are disappeared (.png) |
|--------|
| screen shot 2017-07-03 at 16 47 37    |

| checkboxes with followup questions are visible, others are gone (.gif) |
|--------|
| ie8    |

gulpfile.js
```diff
@@ -8,13 +8,16 @@ var repoRoot = __dirname + '/';
 var npmRoot = repoRoot + 'node_modules';
 var govukToolkitRoot = npmRoot + '/govuk_frontend_toolkit';
 var govukToolkitAppFolder = repoRoot + 'pages_builder/govuk_frontend_toolkit';
+var govukElementsRoot = npmRoot + '/govuk-elements-sass';
+var govukElementsRootFolder = npmRoot = 'pages_builder/govuk-elements-sass'
```
Wynndow: This seems weird to me. Why is `npmRoo`t being reassigned to the same thing as `govukElementsRootFolder`?

Wynndow:
That last commit, to fix that typo has resolved the ie8 issue and the checkboxes now render, see the gif below. Having the extra quote was preventing the `multiple-choice` class being applied correctly. They do use default styles for the checkboxes instead of the new larger styles; do we care about that?

checkboxes-work-in-ie8

gulpfile.js
```diff
@@ -9,7 +9,7 @@ var npmRoot = repoRoot + 'node_modules';
 var govukToolkitRoot = npmRoot + '/govuk_frontend_toolkit';
 var govukToolkitAppFolder = repoRoot + 'pages_builder/govuk_frontend_toolkit';
 var govukElementsRoot = npmRoot + '/govuk-elements-sass';
-var govukElementsRootFolder = npmRoot = 'pages_builder/govuk-elements-sass'
```
allait: Should this be `repoRoot + ...` like `govukToolkitAppFolder` above?

PR #355, opened 2017-07-05
Flash messages universal template

galund:
These came from the buyer app, and there's potential for further refactoring to remove the 'named messages' concept, e.g. the message text should constants that app's `app.main.views.login` module.

Shared template for flash messages, with additional margin.
 - have not ascertained what the effect would be of adding margin to all instances of the banner.

https://trello.com/c/Uak7y047/8-feedback-forms

See also https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/555


risicle:
Hmmm I think it's a little strange to have actual *content* in the frontend toolkit. I'm imagining the code update cycle for this - it means to change a bit of displayed text you need to jump through the whole version bumping process.

But then again, I do see why you did this - it goes some way towards solving the cross-app flash messages problem. *As long as* the apps' frontend-toolkit versions are kept relatively up to date with each other (we're very bad at this from what I remember).

TheDoubleK:
I'm 100% behind the idea of fixing flash message handling, nice one! :clap:

But there are a couple of things here I'm not sure about...

1. I don't see why flash messages should need additional spacing from other banners that get displayed?  If we need it then I'd be tempted to add the margin to the existing banner class rather than have a special "flash messages" class.

2. I don't think that the frontend toolkit is the right place for message content to live.  This is really a place for design patterns, not for words.  I can't see how `app.main.views.login` could be a place for it?  Flash messages are for all sorts of things, not about logging in, and if it's in individual apps then we'd need to replicate all messages across all frontend apps.

The more obvious (to me, anyway) , but also a bit weird place they could go is in `digitalmarketplace-utils`, a constant dictionary, possibly in `digitalmarketplace-utils/dmutils/flask_init.py` or just in a separate file where it can be referenced by the `_base_page` template in each frontend app.

e.g. in Utils you have something like 
```
FLASH_MESSAGES = {
    "password_updated: "You have successfully changed your password.",
    "password_not_updated": "Could not update password due to an error.",
    "supplier-role-required": "You must log in with a supplier account to see this page.",
    ...
    ...
}
```

And in the `base_page` have:
```
{% block flashed_messages %}
  {% for category, message in messages %}
     // Render all the banners
  {% endfor %}
{% endblock %}
```

```

TheDoubleK:
My version would also have the whole issue of bumping utils versions just to bring in new words. But if we're going to try to do things cross-app I'm not sure we can easily avoid that at all.

Another option would be to push them all into the frameworks repo and use the content loader to get at them :trollface:

risicle:
I'm not too sure how scalable it will be to just keep adding the messages to one big if/else statement. I would probably have gone for a one-message-one-template thing.

Does this mechanism have the capability of adding arbitrary payload data to the message, allowing us to do "Successfully deleted opportunity 'Foo Bar'"?

TheDoubleK:
Ooh good call on the customisation @risicle

risicle:
In my mind the cookie should consist of a template label/optional data dict pair. (When I say template I don't *necessarily* mean jinja template, though that would be logical).

Something interesting to *include* would be a simpler fallback static message which the displaying page could show if it didn't recognize the template label.

galund:
Hi both, lots of things for me to reply to here, so apologies if I miss something. Will try to work from easiest to hardest to address...

1. The spacing issue @TheDoubleK is simply because I didn't know how to find and check all the places that used the existing CSS classes. Because I was only going to test the layout of the flash messages, I needed a class just for that. I've absolutely know idea why there wasn't already sufficient margin, but presumably it's been accounted for elsewhere, and I don't really trust the collapsing margins thing enough to change a bunch of layouts that I haven't seen.

2. I don't like having these in the template either. They're only there because I was trying to keep the change minimal in the first instance, and if you don't like it, I'm happy to look at including the next step in this PR.

However, I think we can keep this much simpler than the suggestions here: all of the constants so far came from one file in the buyer app. (Actually, because they're login-related, they belong in a separate app altogether really, but the point stands.) They _don't_ need to be shared between the other apps.

And they don't even need to be in a dictionary, because you can just reference constants when you add the message, rather than lookup names in a dictionary, I think? The template should literally just need to render a string, with some layout adjustment for the message category. This is the standard approach for flash messages, anyway, I'd argue.

There's then also the potential to reduce some duplication in the test rig (which have a secondary copy of every text string).

Definitely if there were any messages that had to be shared between the apps, then dm-utils would be the right place for them, if keeping them in sync seemed like an important constraint in each case. Not sure I know of any messages like that, though.

galund:
Note: there are two messages hard-coded into the templates that are expected to originate from the API.

Personally I think that the API should hold both the human-readable version of the message, and the machine-readable key, but I'm _not_ going to do that refactoring now, so the human-readable one will just have to go into the buyer app (until there's a separate login app).

TheDoubleK:
The point about being shared across applications is that although there are pages we expect users to end up and we think we know what app that will be, in reality the user could be on any page of any of our apps with a flash message set from somewhere else (e.g. user logs in, flash message key gets set in cookie expecting to show up on redirect to a buyer app page, but they go through a tunnel, lose connectivity and hit a bookmark for a page that belongs to the supplier app in the meantime - then that supplier app template ends up trying to render a message from a key that only exists in the buyer app.)

So in reality it would be good if all front-end apps know how to render all possible flash messages.

galund:
Yes, so the 'shared nothing' principle here does suggest making sure all of
the messages are stored only where they originate. (But, as you say, must
then be displayable anywhere.) I'll talk you through my approach Monday.

PR #356, opened 2017-07-13
Fix label in pricing.html

tvararu:
The label for the `maximum_price` field is wrong so when you click on it it selects the `minimum_price` input instead.


tvararu:
Hmm all the other `<label>`s in the page look broken too but that's because of repeating IDs so to fix it for them it's a bit more involved than this.

samuelhwilliams:
Accepted in https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/359, thanks.

Issue #346, opened 2017-05-17
Option select broken for long labels/narrow widths


TheDoubleK:
The label spills out of the option-select container, so words can be cut off or obscured altogether.
e.g. 
![screen shot 2017-05-15 at 09 22 36](https://cloud.githubusercontent.com/assets/6525554/26144700/cac82a1a-3ae1-11e7-8fdf-5260696a8ee7.png)

And here showing the overspilling label:
![screen shot 2017-05-16 at 15 18 45](https://cloud.githubusercontent.com/assets/6525554/26144757/f988e7f4-3ae1-11e7-8caf-47cd088f0860.png)

This is going to become more of an issue with G-Cloud 9 service filters, as many more long filter names are being introduced.

It looks like this is a wider issue with the GOV.UK option select style - e.g. the same problem can be seen on this page: https://www.gov.uk/cma-cases - so if we come up with a fix it would be nice to make a pull-request on the GOV.UK styles at the same time too.

The issue can't easily be seen in our toolkit example page (http://alphagov.github.io/digitalmarketplace-frontend-toolkit/forms/option-select.html) because none of the example labels are long enough to wrap in the full-width option select, but if you tweak the examples to have really long labels you can see it's broken here:
![screen shot 2017-05-17 at 09 24 35](https://cloud.githubusercontent.com/assets/6525554/26144987/b20bcc1a-3ae2-11e7-9726-26fb4903a48e.png)

galund:
deployed to preview already with <https://trello.com/c/IrcO0G3U/491-add-clear-all-filters-link>

PR #357, opened 2017-07-14
Fix text breaking issue with option select boxes

Wynndow:
See this issue [https://github.com/alphagov/digitalmarketplace-frontend-toolkit/issues/346](https://github.com/alphagov/digitalmarketplace-frontend-toolkit/issues/346).

And this trello card [https://trello.com/c/E1WJJbUy](https://trello.com/c/E1WJJbUy)

This seems to work. I'm not sure if this is best way to solve this, or if I should be solving frontend problems in a different way though. Any advice is appreciated. See gifs of different browsers below.

It works with chrome, firefox, safari, IE11. IE8 seems to have a wider issue where it's not responsive. At all. I don't know if this is intentional or not?

### Add `box-sizing: border-box` to options-container label

The labels for options in options containers have their width set to
100% of their parent. They also have a lot more padding on the left than
the right so that they don't clash with the checkbox.

Width doesn't take into account padding, which meant that the actual
label element was extending far too much to the right and so strings
that were long enough to need to be broken would disappear out of the
options container before breaking.

By setting the box-sizing to border-box, the padding is included and so
setting the width of the label to 100% prevents it from spilling out on
the right.

I've also adjusted the width for desktop's to 94% (from 90%). This
reduction in width is to allow for the scroll bar which is only present
when viewing on a desktop width screen. Due to the label now being less
wide that it previously was, leaving a 10% gap for the scroll bar looked
too much. Changing to 94% keeps it very similar to what it was.


samuelhwilliams:
The line separators look like they are terminating really prematurely in this new version. The word wrap is better (insofar as the bug is gone!) but it does leave more whitespace than strictly necessary in some instances. Can we get the separators to go all the way across regardless? Would probably be happy with the early wrapping if the horizontal lines at least don't have gaps in.

## Before <---------------------------- ----------------------------> After
screen shot 2017-07-20 at 11 28 59 screen shot 2017-07-20 at 11 28 46

Wynndow:
I think the gap is there intentionally to allow for the scroll bar. Different browsers have different scroll bar width, hence the large gap I think. What browser was this on? I'm surprised there's such a big difference in the before after. I think it can easily be adjusted though.

samuelhwilliams:
This was on Chrome.

Wynndow:
As discussed, on my machine once you get below 769 pixels wide, the media thing changes and you  the lines go all the way across and there are no scroll bars. If you could replicate the above so we can take a look at it it'd be handy.

Wynndow:
After consulting with @laurenceberry, it's ok to make the lines 100% for desktop widths. It'll just sit behind the scroll bar and look less weird.

samuelhwilliams:
Cool. 👍

Wynndow:
Have removed the width reduction. Here are some images of how it looks in desktop mode.

### Chrome
chrome

### Firefox
firefox

### Safari
safari

### IE11
ie11

### IE8
ie8

PR #358, opened 2017-07-18
Filter on click bugfix

samuelhwilliams:
## Summary
Filter on click results were being cached against the form state when the response was received, not against the form state when the request was sent. This led to the wrong results being displayed if users selected multiple filters quickly (i.e. before the response came back).
Bumps version to 23.0.1

## Ticket
https://trello.com/c/SGxQeBdA/502-filter-on-click


MuriloDalRi:
👍

PR #360, opened 2017-07-20
Adds in CSS for 0 results state suggestions

DilwoarH:
Adds CSS for zero result state suggestion text

Ticket: https://trello.com/c/RDI7f8mf/539-add-zero-results-state


samuelhwilliams:
Will need to bump VERSION.txt here as well. Probably just a patch bump is fine.

PR #361, opened 2017-07-24
Adds in clear-filters.js file which handles clearing of filters.

DilwoarH:
Ticket: https://trello.com/c/IrcO0G3U/491-add-clear-all-filters-link

updates to version number from 23.0.3 to 23.0.4


toolkit/javascripts/clear-filters.js
```diff
@@ -0,0 +1,36 @@
...
+    
+  };
+
+  ClearFilters.prototype.clearFilters = function() {
+    $('.dm-filters input[type=checkbox]:checked')
+      .prop('checked', false)
+      .trigger('change');
+  };
+
+  var clearFilters = new ClearFilters( $('#clearAllFilters') );
```
samuelhwilliams: Not sure why we are passing in the jQuery object here; the called function doesn't use them. Should either not pass anything in, or use what's passed in.

toolkit/javascripts/clear-filters.js
```diff
@@ -0,0 +1,36 @@
+/* 
+  ClearFilters
+
+    Requirements
+      ID: #clearAllFilters
```
samuelhwilliams: Looks like our precedent is to use hyphenation rather than camel case for IDs, so let's change this to `dm-clear-all-filters`.

toolkit/javascripts/clear-filters.js
```diff
@@ -0,0 +1,36 @@
...
+  };
+
+  ClearFilters.prototype.clearFilters = function() {
+    $('.dm-filters input[type=checkbox]:checked')
+      .prop('checked', false)
+      .trigger('change');
+  };
+
+  var clearFilters = new ClearFilters( $('#clearAllFilters') );
+
```
samuelhwilliams: These modules are normally bound to the GOVUK or GOVUK.GDM objects. See other JS files in the frontend-toolkit for examples, but we should probably do the same with this one.

toolkit/javascripts/clear-filters.js
DilwoarH: ops forgot to change, nice spot

toolkit/javascripts/clear-filters.js
DilwoarH: thanks

PR #362, opened 2017-07-25
Adds in Virtual Page view for tracking filters

DilwoarH:
User Story:
As a delivery team I need to understand which filters have been selected so that I can make decisions about which filters to improve or remove.

Virtual Page View class has been refactored to follow JS standards as well as including a new trackFilterAnalytics function.

Ticket: https://trello.com/c/o355GRIi/513-tracking-filters

### DIFF
--- a/VERSION.txt
+++ b/VERSION.txt
@@ -1,5 +1,1 @@
-<<<<<<< HEAD
 23.1.1
-=======
-23.0.5
->>>>>>> updates frontend toolkit to 23.0.5
--- a/VERSION.txt
+++ b/VERSION.txt
@@ -1,1 +1,5 @@
-23.1.0
+<<<<<<< HEAD
+23.1.1
+=======
+23.0.5
+>>>>>>> updates frontend toolkit to 23.0.5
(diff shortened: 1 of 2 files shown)


VERSION.txt
```diff
@@ -1 +1 @@
-23.0.3
+23.0.5
```
DilwoarH: changed to 23.0.5 due to 23.0.4 already in PR

toolkit/javascripts/live-search.js
```diff
@@ -102,6 +103,14 @@ endpoint response (application/json):
     }
   };
 
+  LiveSearch.prototype.filterSelected = function(){
+    if(this.checked) {
+      var key = $(this).attr('name');
+      var value = $(this).attr('value');
+      GOVUK.GDM.analytics.virtualPageViews.trackFilterAnalytics(key,value);
```
samuelhwilliams: This feels like it will need some refactoring. The frontend-toolkit shouldn't have a dependency on a frontend-app. This piece of code is only implemented in the buyer-frontend and is otherwise a hidden bit of code you need to know to add when you want to have a LiveSearch feature.

Can we implement this `trackFilterAnalytics` in the frontend-toolkit?

DilwoarH: I think that's a valid point. We should probably implement this on the toolkit as it can be used generically.

DilwoarH:
Pull Request to toolkit no longer needed.

PR #364, opened 2017-07-28
Fix bad flash message

galund:
Simple change to remove some messages that are not intended for users to see.

https://trello.com/c/4JthSqdx/614-flash-message-problem-on-buyer-account-creation


galund:
I've gone back to 97c32be in the buyer-frontend, and 34adb53 in the frontend-toolkit, and I can't find any evidence that there was a message in this case.

I think there should be, but I'll raise a new ticket for that, as I think it's potentially more complicated than I thought.

PR #365, opened 2017-08-01
Fixes a bug where link remains focused after clicking.

DilwoarH:
Fixes a bug where link remains focused after clicking.

As a buyer
I need to be able to remove all of the filters I have applied quickly
So that I can improve my search results

Ticket:
https://trello.com/c/IrcO0G3U/491-add-clear-all-filters-link


toolkit/javascripts/clear-filters.js
```diff
@@ -23,6 +23,7 @@
     this.$clearFilterLink.on('click', function(event) {
         event.preventDefault();
         _this.clearFilters();
+        $(this).blur();
```
galund: Bit worried about this - what happens if I am a keyboard user, and I want to tick three filters in a row? Does this change mean I have to tab through all the filters again to get back to where I was?

DilwoarH: no - it remains in the same position, ie. when you press tab it will go to the next element after the "Clear filters" button.

PR #366, opened 2017-08-02
Add new optional description div to statistics template

benvand:
* Add new class toolkit/scss/_statistics.scss::statistic-description
* Conditionally add description to toolkit/templates/statistics.html::big_stat if it exists
* Add to pages builder example
* Up the version


Wynndow:
Looks good to me!

PR #367, opened 2017-08-14
Feedback forms

galund:
Styles and JavaScript copied from gov.uk for implementation of the very simple feedback form as seen on the main web site.

It's a shame we can't just fork or even reference the gov.uk equivalents, but: a fork of the entire gov.uk static repo, just for these files, seems like massive overkill; and: referencing their live CSS/JS does not seem sufficiently robust for our needs. If this becomes significant effort to maintain, we can revisit the decision.

This is technically a breaking change, because we are moving some SCSS from one file to another. In practice I don't think there should be any effect on any of our frontends.


toolkit/javascripts/report-a-problem.js
```diff
@@ -0,0 +1,51 @@
...
+    }
+  };
+
+  GOVUK.GDM.ReportAProblem = ReportAProblem;
+
+  ReportAProblem.prototype.trackEvent = function(action){
+    if (GOVUK.analytics && GOVUK.analytics.trackEvent) {
+      GOVUK.analytics.trackEvent('Onsite Feedback', action, { label: '(not set)'} );
+    }
+  };
```
MuriloDalRi: I think we use GOVUK.GDM.analytics so this might not be necessary.

toolkit/javascripts/report-a-problem.js
```diff
@@ -0,0 +1,51 @@
...
+
+  ReportAProblem.prototype.toggleForm = function(evt) {
+    this.$container.toggle();
+
+    if (this.$container.is(':visible')) {
+      this.trackEvent('GOVUK Open Form');
+    } else {
+      this.trackEvent('GOVUK Close Form');
+    }
```
MuriloDalRi: We don't really care about tracking this I think

CHANGELOG.md
```diff
@@ -1,21 +1,37 @@
...
+
+Example app change:
+
+app/assets/scss/application.scss
+```diff
++@import "toolkit/shared_scss/_conditionals2.scss";
+```
+
+## 23.0.0
+
```
samuelhwilliams: Looks like the PR was never included here. It's https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/353. Would be nice to pop it in.

toolkit/javascripts/report-a-problem.js
```diff
@@ -0,0 +1,51 @@
...
+  GOVUK.GDM.ReportAProblem = ReportAProblem;
+
+  ReportAProblem.prototype.trackEvent = function(action){
+    if (GOVUK.analytics && GOVUK.analytics.trackEvent) {
+      GOVUK.analytics.trackEvent('Onsite Feedback', action, { label: '(not set)'} );
+    }
+  };
+
+  $(document).ready(function() {
+    new GOVUK.GDM.ReportAProblem($('.report-a-problem-container'));
```
samuelhwilliams: If you pop this constructor in the init function, does it work as intended? I think that's what it's for, and nothing else we have hooks into document ready, so would be good to be consistent if it works.

toolkit/templates/report-a-problem.html
```diff
@@ -0,0 +1,22 @@
+<div class="report-a-problem-container">
+  <div class="report-a-problem-inner">
+    <div class="report-a-problem-content">
+      <h2>Help us improve the Digital Marketplace</h2>
+      <form autocomplete="off" action="/feedback" method="post">
+        <input id="uri" name="uri" type="hidden" value="{{ request.url }}">
+
+        <p>Please don’t include personal details or sensitive information about your business or project.</p>
```
samuelhwilliams: Project???

galund: lol. @karlchillmaid ?

karlchillmaid: 🙄 
I'd just lose "or project"

galund: Does that sufficiently cover government users / civil servants? If so that's fine :)

karlchillmaid: Good point. To be safe we could say "Please do not include personal details or sensitive information about your company or department."
Whatever we do, we should definitely change "don't" to "do not".

MuriloDalRi:
👍

PR #375, opened 2017-09-22
Form macros: multiple improvements

risicle:
Related to https://trello.com/c/iaRQlj6y/87-toolkit-forms-macros-are-different-in-all-of-the-apps and https://trello.com/c/W3d7VMrO/72-supplier-app-use-frontend-toolkit-macros-3

A few things done here:

* allow context vars to be overridden by macro `kwargs`: this allows apps which need small alterations in the handling of `question_content` to use these macros.
* reordered argument lines alphabetically: this makes it clearer what arguments are present & missing from each invocation.
* added some parameters extracted from `question_content` which had been missing.

I found it hard to decide whether to make this a minor or major version bump seeing as it could hypothetically have unexpected results. In the end I decided to make it minor, because I wouldn't have been able to think of anything specific enough to put in the changelog. "Check everything"? I would have hoped tests would catch any problems.

The only way I could see things being affected is is a question defined e.g. `question_content.unit` but for some reason deliberately didn't want it expressed.

I'm aware that this touches almost every line here, but I've done this as a number of discrete commits, each performing one simple operation that can be checked for equivalence if you like.


TheDoubleK:
Some of the new possible overrides are kind-of weird to have at all - value, error, type, and probably some others too I can't imagine why you would ever legitimately want to override - but I guess it makes sense for consistency to allow it for all params.

toolkit/templates/forms/macros/forms.html
```diff
@@ -193,16 +197,16 @@
...
+    hint=kwargs.get("hint", question_content.hint),
+    link=kwargs.get("link", question_content.link),
+    message=kwargs.message,
+    name=kwargs.get("name", question_content.id),
+    optional=kwargs.get("optional", question_content.optional),
+    question=kwargs.get("question", question_content.question),
+    question_advice=kwargs.get("question_advice", question_content.question_advice),
     question_number=kwargs.question_number,
-    message=kwargs.message
+    value=kwargs.get("value", "Document uploaded {}".format(data[question_content.id]) if data[question_content.id] else None)
```
TheDoubleK: Oooh this is weird.  But it's what was there already...

It's different in supplier frontend, so calls there will need to override the value as a kwarg when making the switch to use these.

risicle: Yup & it does. https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/748/files#diff-712f48621dd493057d3678f370d01074L157

toolkit/templates/forms/macros/forms.html
```diff
@@ -211,20 +215,20 @@
...
+    hidden=kwargs.get("hidden", question_content.hidden),
+    hint=kwargs.get("hint", question_content.hint),
+    message=kwargs.message,
+    name=kwargs.get("name", question_content.id),
+    optional=kwargs.get("optional", question_content.optional),
+    question=kwargs.get("question", question_content.question),
+    question_advice=kwargs.get("question_advice", question_content.question_advice),
     question_number=kwargs.question_number,
-    message=kwargs.message
+    smaller=kwargs.get("smaller", True),
```
TheDoubleK: This `smaller` is another one I'm not sure should be possible to override - this enforced `smaller` is pretty much what distinguishes this `number` macro from a regular textbox macro I think.

TheDoubleK:
I think you're right about the versioning... any code already using these macros should not break as far as I can tell.

The potential issues arise when switching out local macros for these toolkit macros, but that's a thing that needs to be done with care whatever, and isn't a "break" in the toolkit.

toolkit/templates/forms/macros/forms.html
risicle: I don't really feel it's my job here to enforce arbitrary restrictions. Any restrictions I put on I can see having to be overridden and it creating an extra piece of work to come here & change them.

galund:
Can you give an example of what you couldn't do before, but you can do now -- because I don't really understand what this is about.

risicle:
@galund see https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/748 where most of the macros want to e.g. override the `question` arg with one piped through `question_references`. The previous construction here would have happily ignored that, with the macro's setting for `question` getting the final say, meaning the supplier app had to keep a modified copy of the macros to get the right effect.

galund:
Hmm, I still don't get it. The macros are of dubious use in the first place, and if you're literally overriding the one thing they do, then why bother at all? Just call the 'include' directly `with` whatever arguments you want, and skip the macros altogether?

risicle:
I am *trying* to retain compatibility for all consumers of the toolkit. There is *one* useful effect that the macros give us that you don't get just using the raw includes and that is the body of the macro supposedly starts with a clean context. It's possible to "call" an `{% include` thinking you're omitting arguments and thereby signifying the defaults, but in the background of your template you have stray context variables colliding with the fairly common names like `value`, `hidden`, `message`.

galund:
FWIW I think that more complex substitution of defaults etc should be done outside the template, so that the notional "template designer" person doesn't have to worry about all this stuff. This might be achievable with the existing classes we have (i.e. creating new instances of them and having a special factory method), or creating new classes with the necessary compatibles interfaces with properties that default appropriately.

This could make the macros simpler, even if we still feel we need to keep them because of the context issues? It just feels like we're trying to do too much in the templating layer (in general, though these macros are the example that keeps coming up).

risicle:
I don't really disagree with you.

PR #376, opened 2017-09-22
736 bug styling on dos opportunities page

galund:
Apply style based on more generic class name - should make sense in the context of https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/616

https://trello.com/c/DG6GGfib/736-bug-styling-on-dos-opportunities-page


toolkit/scss/search/_search-summary.scss
```diff
@@ -13,7 +13,7 @@
   }
 }
 
-#js-dm-live-search-wrapper .search-summary .search-summary-count {
+.search-results-page .search-summary .search-summary-count {
```
idavidmcdonald: The word `page` feels so generic and something that would be added to all the content on a page yet this is not being added to the whole page in the buyer app, more just the content or main section of the page.

The best I could think of that maybe slightly better would be `wrapper`, `content` or even just `search-results`.
 I don't know if these are any better though so happy to stick with what you have suggested unless this spurs you to change things.

galund: Point taken. It's a fairly top-level thing, so I think page is probably ok. Wrapper is even more generic if anything, and having something that implies you're looking for an element somewhere near the top of the containment hierarchy is vaguely helpful? I'll stick with this for now but reserve the right to be wrong and change again in future :-)

idavidmcdonald: Seems reasonable, all good on my side!

PR #377, opened 2017-09-26
Feedback forms fix

galund:
Yep it was an easy fix, and it seems to still work when logged-out too.

https://trello.com/c/Uak7y047/706-feedback-forms


galund:
Yup, so it is a function, and the function is called `generate_csrf` as it happens.

It gets put into the template with the name `csrf_token` by `flask_wtf.csrf.CsrfProtect.init_app`, and while I agree with you that they've named it wrong (if it needs to be called like a function, that ought to be obvious from the name), it's not really something we can change, as that's a shared library.

CrystalPea:
Ahh, thanks for that exquisite explanation! ❤️ 
All good then 👍

PR #380, opened 2017-10-03
Run the toolkit with Python 3

TheDoubleK:
I thought I had a fix to add to the toolkit but turns out it's OK.

In the meantime I got it running locally under Python 3, and this was the minimal set of changes I had to make so that `make serve_pages` and `make watch_for_changes` both work OK.

This new version will no longer work with Python 2, but I think that's what we want these days?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@ SHELL := /bin/bash
 VIRTUALENV_ROOT := $(shell [ -z $$VIRTUAL_ENV ] && echo $$(pwd)/venv || echo $$VIRTUAL_ENV)
 
 virtualenv:
-	[ -z $$VIRTUAL_ENV ] && [ ! -d venv ] && virtualenv venv || true
+	[ -z $$VIRTUAL_ENV ] && [ ! -d venv ] && virtualenv -p python3 venv || true
 
 requirements: virtualenv
 	${VIRTUALENV_ROOT}/bin/pip install -r pages_builder/requirements.txt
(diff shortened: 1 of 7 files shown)


benvand:
LGTM 👍  Standard py2 > py3 stuff. 
Could do a version bump but this could just as easily go along with the next version since it doesn't break anything or add functionality

Wynndow:
I can't comment on it here, but it might be good to update the makefile to use python3 for the virtualenv? line 5 would be:
`[ -z $$VIRTUAL_ENV ] && [ ! -d venv ] && virtualenv -p python3 venv || true
`

benvand:
@Wynndow 
Found a nifty python3 specific way to do this the other day:
`[ -z $$VIRTUAL_ENV ] && [ ! -d venv ] && python3 -m venv venv`

benvand:
Although saying that we're not using that anywhere else so... Up to you @TheDoubleK

galund:
Yup agree the way @benvand suggests is right - `virtualenv` isn't supposed to be a thing on Python 3 any more.

Wynndow:
Nice!

TheDoubleK:
"virtualenv isn't supposed to be a thing on Python 3 any more." 

So... what is the `venv` thing?

TheDoubleK:
Damn, hit the wrong button.

TheDoubleK:
OK with Ben's thing when I try to `make serve_pages` I just get 
```
[ -z $VIRTUAL_ENV ] && [ ! -d venv ] && python3 -m venv venv
make: *** [virtualenv] Error 1
```

So I'm going to go with Chris's version instead.

TheDoubleK:
Yeah I don't think this merits a version bump as the toolkit itself is unchanged.

galund:
@TheDoubleK venv is just part of python now, PEP405 justifies it along the lines of "we can do this without so many hacks because we are inside Python" which makes sense I think. https://docs.python.org/3/library/venv.html

PR #384, opened 2017-10-23
Lists and paragraphs inside warning notification banner

katstevens:
Required for this Trello ticket https://trello.com/c/7kq7pSTr/83-allow-buyers-withdraw-open-opportunities

warning-banner-with-list-para

The notification banner macro template now accepts a `content` parameter (displaying black text) as an alternative to the `message` parameter (which will display red text as usual, i.e. this should not introduce any breaking changes). 

I wasn't sure whether to apply the new styles for just the `.banner-destructive-with-action` class (i.e. the only place it's going to be used so far), or for all the banner classes (so we can use it elsewhere in the future). Advice welcome!


TheDoubleK:
I reckon allow the new parameter and add the associated style for all three of this-style of banner ('success', 'warning', and 'destructive').

Now we have it, everyone will want it, and looking at the patterns page I'd expect all three of those to behave exactly the same way bar the colour.

TheDoubleK:
"temporary-message" is a different kettle of fish.

TheDoubleK:
Actually, given that all banner types - including "temporary message" - will accept and display the new `content` bit then maybe add the styles for all (or just once at the `.banner` level)?

katstevens:
Yep, that sounds reasonable. I can leave out the font colour style for the temporary message (blue background).

TheDoubleK:
Or just put it in for all and when someone comes to actually use it in temporary message they can override the style with what they want.

katstevens:
temp-message-banner

Getting the temporary-message to show white text was straightforward in the end. Now all our banners can use `content` to show lists 🎉

PR #385, opened 2017-11-02
Pluralize validation errors lede

samuelhwilliams:
## Summary
Pluralize validation errors lede


katstevens:
Nice! Although given this might break a bunch of test assertions, should the version bump be higher?

samuelhwilliams:
Not sure. Good point.

Gut feeling says that we shouldn't make it a major version bump if it is just breaking things on the dev side (i.e. tests).

Open to a good reason to the contrary though.

TheDoubleK:
Did this come from a content person?  I think the existing message was designed to work for multiple errors and we were happy with that.

samuelhwilliams:
@TheDoubleK No, it hasn't come through content. The content already existed on whichever PR it was you were just working on, so didn't think it needed it. It's also just more grammatically correct.

Happy to get input though?

TheDoubleK:
I think the pre-existing message is really old content that had been superseded by the one in the template everywhere except admin, so I'm not sure we want it any more.  I think it's fine to have the singular-style message here followed by a list, and the current message was content designed to work for one or more errors on the page.

But I might be wrong, so could be worth running by a content person.  I'm sure we're not particularly attached to the old admin message though.

TheDoubleK:
Maybe I'm over thinking this and we should just do it though.

samuelhwilliams:
@karlchillmaid @constancecerf Do you want to give input or whether this should be pluralised or not, when we are listing multiple errors?

Example of when this might be relevant would be in the 'After' image here: https://github.com/alphagov/digitalmarketplace-admin-frontend/pull/303

samuelhwilliams:
No real pull from content that we should do this, so going to just close it.

PR #386, opened 2017-11-03
Adds content support for instruction lists

DilwoarH:
REQUIRED BY: https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/660

The space between the numeral and the title seems to vary on different items.

1 and 7 look ok, the others look too close together.

Ticket: https://trello.com/c/LHDlR93V/123-bug-spacing-issue-on-the-compare-and-decide-saved-search-task-list-page

## Before:
screen_shot_2017-10-20_at_10 09 44

## After:
screen shot 2017-10-24 at 11 53 10

# Documentation Update
screen shot 2017-11-07 at 14 18 38


toolkit/scss/_page-headings.scss
```diff
@@ -22,6 +22,10 @@
   }
 }
 
+%page-heading-h2 {
+  @include bold-27;
```
DilwoarH: don't use  - take it to buyer frontend

toolkit/templates/instruction-list.html
```diff
@@ -64,6 +64,23 @@ <h2 class="instruction-list-item-body">{{ item.body }}</h2>
...
           {% endfor %}
         {% endif %}
+
+        {% if item.content %}
+          {% for content in item.content %}
+            {% if content.type == 'text' %}
+            <p class="instruction-list-item-top{% for class in content.class %} {{ class }}{% endfor %}">{{ content.text|safe }}</p>
+            {% elif content.type == 'action' %}
+            <a class="instruction-list-item-action button-save" href="{{ content.href }}" 
+            data-analytics="trackEvent"
```
DilwoarH: need to put this around if statements

VERSION.txt
```diff
@@ -1 +1 @@
-24.4.4
+25.0.0
```
galund: If this is a major (breaking) change then you'd need to update the changelog - but I don't think it breaks anything, so the new version should be `24.5.0`.

toolkit/templates/instruction-list.html
```diff
@@ -64,6 +64,24 @@ <h2 class="instruction-list-item-body">{{ item.body }}</h2>
             {% endwith %}
           {% endfor %}
         {% endif %}
+
+        {% if item.content %}
+          {% for content in item.content %}
```
galund: The naming here is a bit yuck. Is there any possibility of changing at this point? Having content as a _list_ containing content as a _dict_ means you never really know what _content_ actually means!

toolkit/templates/instruction-list.html
```diff
@@ -64,6 +64,24 @@ <h2 class="instruction-list-item-body">{{ item.body }}</h2>
             {% endwith %}
           {% endfor %}
         {% endif %}
+
+        {% if item.content %}
```
Wynndow: Having `item.content` and `item.body` is quite confusing, they both imply the same thing. It would be preferable if `item.body` was called something else I think. Header or something. Although that's going to be a breaking change.

DilwoarH: item.body is the heading - the whole template itself has misleading property names. We should think about refactoring.

Wynndow: How much work would it be to change? If we made the breaking change here, it'd only be the buyer-frontend that needs fixing now. Other apps can be updated as they pull in the change.

galund: (If we did do something like that, which sounds a good idea to me, then we'd need to make sure the failure mode was super-clear. As in, we should make sure that the functional tests _do_ break, so that it wouldn't be possible to upgrade without fixing by mistake. If for some reason that wasn't straightforward then we could always rename the macro :) )

DilwoarH: I feel like that specific change would need to go through the tech debt process and is out of scope for this particular issue.

DilwoarH:
@galund @Wynndow - I have updated the page with the requested changes.

toolkit/templates/instruction-list.html
Wynndow: I dunno, it feels like we're adding the `content` attribute of the item here which is causing the confusion, so this might be the time to look at it. I've just had a look and this is the only place in the buyer frontend that is using an instruction-list, so should be pretty easy to fix.

Wynndow:
@DilwoarH Could you add an example to the pages builder?

toolkit/templates/instruction-list.html
```diff
@@ -64,20 +64,36 @@ <h2 class="instruction-list-item-body">{{ item.body }}</h2>
             {% endwith %}
           {% endfor %}
         {% endif %}
+        {% if item.content %}
+          {% for item_content in item.content %}
```
galund: I'm not sure this is any better, sorry! - you've still overloaded the word 'content' to sometimes be a dictionary, and sometimes be a list.

I think it might be a lot simpler (assuming we keep this template) if item.content was just a blob of markup that got passed in, rather than having list items with various types at all? It'd be a lot more flexible if we want to have various things in each step in future.

I get that along with the types there are a bunch of classes  i.e. instruction-list-item-top etc.

Is there actually CSS defined for those in the frontend-toolkit?

DilwoarH:
@Wynndow - I have added in the documentation

Wynndow:
Pages builder looks good. I'm still wondering about making the breaking change thought to make the template less confusing.

DilwoarH:
@galund - I have updated the variable name

toolkit/templates/instruction-list.html
```diff
@@ -64,20 +64,36 @@ <h2 class="instruction-list-item-body">{{ item.body }}</h2>
...
-        <div class="notice">
-          <i class="icon icon-important">
-            <span class="visually-hidden">Warning</span>
-          </i>
-          <strong class="bold-small">
-            {{ item.important }}
-          </strong>
-        </div>
+          <div class="notice">
+            <i class="icon icon-important">
```
galund: FWIW I think this should have been `<em>`

DilwoarH: the Icon tag?

galund: Well, the content is the word "Warning", so it's reasonable to assume that any italics that got rendered were for the purpose of emphasis, rather than just cosmetic. But who knows what would break if you touched that, so don't :)

PR #387, opened 2017-11-09
Analytics attributes for summary table top links

katstevens:
Part of this ticket: https://trello.com/c/kQpbazwt/85-allow-suppliers-edit-submitted-application-if-opportunity-is-still-open

We want to fire GA tracking events for summary table links in the Brief Responses FE (see https://github.com/alphagov/digitalmarketplace-brief-responses-frontend/pull/11). Adding optional `category`, `action` and `label` attributes on the links means we can avoid a lot of unnecessary javascript.  

The toolkit currently uses the same attributes for the document download template (see `documents.html`) - the `top_link` macro uses exactly the same naming, for consistency.

Example usage:
```
{% with
    label="Edit",
    url=url_for('.edit_brief_response', brief_id=brief.id, brief_response_id=brief_response.id, question_id=item.id, edit="edit"),
    analytics="trackEvent",
    category="internal-link",
    action="Edit Supplier Application",
    analytics_label=brief_response.status
%}
   {{ summary.top_link(label, url, analytics=analytics, analytics_category=category, analytics_action=action, analytics_label=analytics_label) }}
{% endwith %}
```

Would give:
```
<a href="/suppliers/opportunities/5197/responses/29422/dayRate/edit" data-analytics="trackEvent" data-analytics-category="internal-link" data-analytics-action="Edit Supplier Application" data-analytics-label="submitted">Edit</a>
```


DilwoarH:
hey guys - do you think any of the documentation needs updating?

katstevens:
We were wondering about that - there are no visible differences to the links so it seemed a bit pointless to add an example to the pages_builder. Is there anything documented for the `documents.html` analytics attributes that we can add to?

TheDoubleK:
There's a page in the manual about how we implement analytics - https://alphagov.github.io/digitalmarketplace-manual/google-analytics.html

Maybe worth a paragraph here that it can be done using the toolkit?

TheDoubleK:
It's nice to be able to do this with the toolkit, but seems odd to add this to just one type of template. But I'm not saying add it to everything right now.

I don't really understand why we care when people edited their application (i.e. before or after submitted)?  I also don't see why we can't get this from our URL analytics - if we filter the visited URLs by brief ID we'll clearly be able to see the submission event and what happened before and after it.  So I'd probably argue against having these events at all.

Unless I've missed something about what they're supposed to tell us?

katstevens:
Unfortunately the final page of the flow doesn't include the Brief Response ID in the URL, so we can't see if a specific supplier has finished the flow or not. Tobi, Ash and I talked about this in detail - the user need is to track whether suppliers are using the new edit feature (we're not bothered about tracking an individual supplier's journey).

katstevens:
A few lines added to the manual here: https://github.com/alphagov/digitalmarketplace-manual/pull/35

As this branch is out of date I'll rebase and bump the version number.

katstevens:
Yep! Thanks Kev 😸

PR #388, opened 2017-11-21
Notification banner fix

katstevens:
Fixes bug where the notification banner template was unintentionally using a `content` context variable in the opportunity page (in the Buyer FE - see below). The variable now has a more specific name 😸 

briefs-banner-bug


benvand:
Nice 👍

PR #389, opened 2017-12-05
Extends filterSelected event to allow multiple frameworks

DilwoarH:
Extends filterSelected event to allow multiple frameworks

This work was done to meet the acceptance criteria:
`It should send the page view to Google Analytics in the following format: /{framework family}/{lot}/filters/{group}/{key}/{value}/vpv`

Ticket: https://trello.com/c/g0h1Ibq6/137-how-to-measure-impact-of-adding-filters-to-dos-opportunities-page


toolkit/templates/forms/option-select.html
```diff
@@ -2,7 +2,7 @@
   <div class="container-head js-container-head">
     <div class="option-select-label">{{ label }}</div>
   </div>
-  <div class="options-container" id="{{ options_container_id }}">
+  <div class="options-container" id="{{ options_container_id }}" data-framework="{{ framework_family }}" data-current-lot="{% if current_lot %}{{ current_lot.slug }}{% endif %}">
```
Wynndow: What if there's no `framework_family`? If an option select is used for something not framework related, you'll get something weird here?

DilwoarH: Good point, I'll get that updated.

PR #390, opened 2017-12-06
Users should by able to see the full list of filters

DilwoarH:
## Users should by able to see the full list of filters

User story
----------------------
As a user, when I'm searching through a list
I need to be able to see all the filters that I have selected
So that I can tell how I have refined my search results

**The people required for kick off are:**

- Laura + developer + Jeremy

**Context/Problem**

The height of the filter boxes is fixed and this means if the list of filters is above 6, the filters are hidden and users have to scroll down to see the full list. On the DOS opportunities page there is no search summary so if users select filters they won't be able to refer back to which ones they have selected easily.

**What are we making?**

- changing the height of the filter boxes so it is no longer restricted and  expands to include the full range of filters
- filters already behave this way on mobile

**Info/Consideration**
- @jeremy will update the screenshots of the user flow

Ticket: https://trello.com/c/9oSx43Mv/219-users-should-by-able-to-see-the-full-list-of-filters

Before:
dec-06-2017 14-47-56

After:
dec-06-2017 14-48-36


Wynndow:
Is this going to affect all option selects on the digital marketplace? Do we use them anywhere else other than filtering search results?

DilwoarH:
@Wynndow - yes this would affect all option boxes but it is only used in filters template

screen shot 2017-12-08 at 10 18 05

Wynndow:
How does the G-Cloud search page look? Cloud hosting has a bunch of lot specific filters. This could make that 1/3 column look a lot longer? Or are we happy with that?

DilwoarH:
I think that's the desired behaviour now.

PR #393, opened 2017-12-14
Copy h2 'heading-large' style from GOVUK elements

katstevens:
Done as part of this ticket: https://trello.com/c/wxCSXbeH/159-update-homepage-group-framework-tasks-by-framework-and-add-headings

Until now we haven't needed the larger `h2` style - now we do (in the Admin home page). I've copied exactly what we've done for `heading-xmedium`.


toolkit/scss/shared_scss/_dmspeak.scss
```diff
@@ -20,6 +20,10 @@
     @extend %heading-small;
   }
 
+  .heading-large {
```
TheDoubleK: I'd be tempted to order these small, medium, large in the file.  Or large, medium small.  But not medium, small, large.

toolkit/scss/shared_placeholders/_dm-typography.scss
```diff
@@ -29,3 +29,15 @@
   }
 
 }
+
+// Cut and pasted from the `.heading-large` styles in the `_elements-typography.scss` file in GOVUK Elements:
```
TheDoubleK: Same here with ordering.

samuelhwilliams:
I don't think I understand the need for this. Shouldn't all our headings at a given level be as 'important' as each other?

katstevens:
The plan is to use `<h2 class="heading-xlarge">` and `<h3 class="heading-xmedium">`.

katstevens:
> grammer

Sorry, this was a terrible Molesworth joke. I can squash 😸

toolkit/scss/shared_scss/_dmspeak.scss
katstevens: Good spot - now fixed.

samuelhwilliams:
Oh I see. It seems weird to me that we don't just define h1/h2/h3 directly as having certain styles... but I guess we don't, so this must be fine. 👍

TheDoubleK:
Sorry for not getting the joke... it would have been fine to leave it in :)

There is a part of me that feels we should have basic styles defined for H1, H2, H3 too.

PR #396, opened 2018-01-08
Remove IE6 specific styles

MuriloDalRi:
https://trello.com/c/06BhH9ow/262-confirm-which-browsers-we-are-supporting-and-document

### DIFF
--- a/toolkit/scss/README.md
+++ b/toolkit/scss/README.md
@@ -9,8 +9,8 @@ Support for older versions of Internet Explorer is made possible by the use of t
 [_conditionals.scss](../govuk_frontend_toolkit/stylesheets/_conditionals.scss).
 
 ```
-@include ie(6) {
-  /* Block of IE6-specific SCSS */
+@include ie(7) {
+  /* Block of IE7-specific SCSS */
 }
 ```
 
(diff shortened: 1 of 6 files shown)


idavidmcdonald:
https://github.com/alphagov/digitalmarketplace-frontend-toolkit/blob/434ad307913651ecb041ab94bdee748ebe066d1a/toolkit/scss/README.md

I'm looking at the scss readme which looks like it could also get updated. IE6 is mentioned and there is a broken link at the moment.

Just having a review to see if I can find any other changes no longer needed.

idavidmcdonald:
https://github.com/alphagov/digitalmarketplace-frontend-toolkit/blob/a1693c7e0ffabe159d1d5cfaed38bf82920ef2b6/toolkit/scss/forms/_summary.scss#L63

This is also something worth looking into if we can remove it or not (at a glance it looks like yes but I haven't taken the time to fully understand it to make sure it is).

idavidmcdonald:
https://github.com/alphagov/digitalmarketplace-frontend-toolkit/blob/36cd6177319c6b16473cc01c130f4206c159cbab/toolkit/scss/forms/_pricing.scss#L41

MuriloDalRi:
I didn't know about that README file! That should really be part of the root dir one

idavidmcdonald:
Yeah I think I ran into problems with that before - I don't know if I like nested READMEs.

Feel free to move it to the main one and tidy up if you'd like.

katstevens:
Definitely 👎 to nested READMEs - no-one will ever find them.

MuriloDalRi:
There's a link to that file on the main README so I guess it's not _that_ bad

toolkit/scss/README.md
```diff
@@ -27,14 +27,14 @@ stylesheets are wrapped in conditional comments to ensure that:
...
+The SCSS file for IE7 just includes `forms/index.scss` while setting a
 few variables. The mixins from `_conditionals.scss` use these variables
 to determine whether or not to insert the block of browser-specific CSS.
 
 ```
 $is-ie: true;
-$ie-version: 6;
-$mobile-ie6: false;
+$ie-version: 7;
+$mobile-ie7: false;
```
idavidmcdonald: This is worth double checking as it might not actually be a flag.

samuelhwilliams:
@MuriloDalRi if this is still good, let's get it in. 👍

PR #400, opened 2018-01-25
add optional aria-controls attribute to most form inputs

risicle:
Controlled by aria_controls context variable

Trello: https://trello.com/c/qUViIfu9/172-dynamic-content-on-the-search-pages-which-updates-without-a-page-reload-is-marked-up-using-wai-aria-live-regions

I've been quite over-generous on which elements to add this ability to, even ones which aren't really likely to be used to trigger live updates, because I'm favouring symmetry over special-case-ism here.


risicle:
Bah yes

katstevens:
All looks good - I grepped for `<input..` and there are a few more lurking in the toolkit that aren't covered in this PR, but I don't know if they all need ARIA stuff? e.g. I don't think the `report-a-problem` template is going to affect live areas but maybe `button.html` will?

risicle:
Ah yes I didn't see `button.html` because it wasn't in the `forms` directory.

PR #401, opened 2018-01-29
search: add .search-summary-accessible-hint-wrapper style

risicle:
Trello https://trello.com/c/qUViIfu9/172-dynamic-content-on-the-search-pages-which-updates-without-a-page-reload-is-marked-up-using-wai-aria-live-regions

This style is needed to visually "hide" the summary hint which is to be read out.


toolkit/scss/search/_search-summary.scss
```diff
@@ -28,4 +28,8 @@
   @include media(tablet) {
     @include core-16;
   }
-}
\ No newline at end of file
+}
+
+.search-summary-accessible-hint-wrapper {
```
galund: My only complaint is that the word 'wrapper' means pretty much nothing. If most-things-like-this-thing get the word "wrapper" on the end, then it's fine I guess.

risicle: It's because it's the level *up* from the `.search-summary-accessible-hint`

risicle:
Oh dear I need to bump versions here too.

brenetic:
🚀 SLGTM

PR #402, opened 2018-02-07
summary table styles: move `text-align: left` rule to actual th element

risicle:
Trello https://trello.com/c/sFnBpHvr/120-there-is-a-table-heading-alignment-issue-on-ie11-or-below-browsers

Expecting it to be inherited from the grandparent `<thead>` doesn't appear to be reliable - certainly not on IE where the default for `<th>`s happens to be `text-align: center`.

Before:
ie11_summary_table_wrong

Now: 
20180207_ie11_summary_table


toolkit/scss/forms/_summary.scss
```diff
@@ -269,13 +269,13 @@ table.summary-item-body tbody .summary-item-field-first-narrow-no-padding {
 
 .summary-item-field-headings-visible {
```
katstevens: I'm curious as to where we actually use this class (plenty of usage of the one below). I wonder if we can delete it altogether?

risicle: This is the class that gets applied to the `<thead>` itself.

Hmm.... but yes, I do suspect that `border-top` has no effect on `<thead>`. Or at least it *shouldn't*. Who knows about weird-browser-support.

PR #404, opened 2018-02-14
Move supplier macros to tookit

Wynndow:
There are some macros in the supplier frontend which wrap the toolkit form templates.

We already have some macros in the toolkit which do a similar thing but are designed to
use question objects generated by the content loader.

The macros defined in the supplier frontend are useful for when we don't have question objects, but still want a convenient way to generate inputs.

Moving them to the toolkit allow them to be used by other apps if desired.


toolkit/templates/forms/macros/forms.html
```diff
@@ -321,3 +321,34 @@ <h2>{{ question_content.question }}</h2>
     {% endfor %}
   </fieldset>
 {%- endmacro %}
+
+{% macro question_input(name, question, value='', hint='', errors=None) -%}
```
samuelhwilliams: This should maybe be called `question_textbox` to be a bit more symmetrical with `question_textarea`.

Wynndow: I'll take it!

PR #406, opened 2018-02-20
58 summary table top link accessibility

katstevens:
Trello: https://trello.com/c/sYJYKnK9/58-how-summary-tables-handle-their-links-means-pages-can-contain-multiple-links-with-duplicate-text

Adds `<span class="visually-hidden"> {{ hidden_text }}</span>` to the summary table top link, so that screen readers can differentiate between the 'Edit' links for each table. This option was chosen over adding an `aria-label="My extra text"` as per guidance from the accessibility team ('don't use ARIA unless you have to').

While trying to test this out, I managed to also fix the documentation build issue: https://trello.com/c/8xMrtTs0/346-fix-jenkin-job-for-frontend-toolkit-documentation


pages_builder/pages/proposition-header.yml
```diff
@@ -2,6 +2,7 @@ pageTitle: Page header
 assetPath: govuk_template/assets/
 examples:
   - |
+    {% macro url_for(mocked_link_for_documentation_only) %}#{% endmacro %}
```
DilwoarH: can this be done globally once?

katstevens: Each example is rendered separately. We could potentially do it in the base template but I think we'd have to import the macro into the example again anyway, and at least here it's close to the code that's using it.

DilwoarH:
LGTM 👍 

Just one thing before someone else says it, it might be worth just squashing all the commits together 🌚

katstevens:
Haha, I was keeping them separate as one's a fix and one's a feature 😸 Do we still need to keep the version bump as a separate commit on this repo? The README implies that we do...

PR #408, opened 2018-02-21
Common analytics JS and custom dimension meta tags

brenetic:
This is the proposed approach to DRY'ing the GA related code. We have left `_events.js` out as we imagine this to be app specific. There may well be others as we progress.

https://trello.com/c/Jj4cz4u4/336-implement-a-custom-dimension-for-buyer-supplier-and-admin-in-ga

Paired with @katstevens 

**Update 23/2/18:**
This PR now covers
- Copying the common analytics JS files from the FE apps to the toolkit
- `_events.js` will be kept in the FE apps for app-specific events 
- Adding a new `custom-dimension.html` template to the toolkit, for setting custom GA dimensions where needed in the FE apps. For usage see the example included in `pages_builder`.

What this does *not* cover
- Refactoring the analytics code/structure itself
- Moving JS analytics tests into the toolkit


toolkit/javascripts/analytics/_init.js
```diff
@@ -0,0 +1,10 @@
+(function(root) {
+  "use strict";
+  // We've copied this from the Briefs FE
```
galund: This comment won't make sense as soon as the briefs PR lands, so it may as well not be there :)

katstevens: Yep, that was just to stop us getting mixed up with the duplicate code in the FE apps - that will get taken out before merging 😸

toolkit/javascripts/analytics/_init.js
```diff
@@ -0,0 +1,10 @@
...
+  root.GOVUK.GDM.analytics.init = function () {
+    this.register();
```
galund: There's something wrong here about the layout, although I get that we're just moving the existing thang... might it make sense to have the `// include` things here, so that you only have to include `_init.js` and then you get the rest? (In fact, at that point maybe you should lose the underscore - we should be finding a way to indicate _something_ as the entry point for this code?)

katstevens:
This should hopefully not require a major version bump now:
- custom meta template can be pulled into FE apps whenever it is needed
- the 'new' analytics files can be linked in the FE app's `/app/assets/javascripts/_analytics.js` whenever they're ready to be used (and the old assets analytics files deleted)

toolkit/javascripts/analytics/_init.js
katstevens: Yep, that makes sense to me.

katstevens: I'm tempted to keep the underscore though, as the FE apps have everything underscored apart from the file we actually end up serving (`application.js`)

pages_builder/pages/custom-dimensions.yml
```diff
@@ -0,0 +1,15 @@
+pageTitle: Custom dimensions
+assetPath: govuk_template/assets/
+pageDescription: Meta tags for Google Analytics custom dimensionsto be included in the <head> tag.
```
galund: missing space

pages_builder/pages/index.yml
```diff
@@ -15,6 +15,7 @@ content: >
         <li><a href="browse-list.html">Browse list</a></li>
         <li><a href="button.html">Buttons</a></li>
         <li><a href="contact-details.html">Contact details</a></li>
+        <li><a href="custom-dimensions.html">Custom dimensions</a></li>
```
galund: Is there actually any point in having this in the pages builder? - there's nothing visible to actually preview?

toolkit/templates/custom-dimensions.html
```diff
@@ -0,0 +1,9 @@
+{# User-level custom dimensions #}
+{% if current_user.role %}
+    <meta name="ga_customDimension" data-id="10" data-value="{{ current_user.role }}">
```
galund: 2-space indents for template files

pages_builder/pages/index.yml
katstevens: I quite like having the usage example on there and seeing what tags get rendered. It's minimal at the moment but when we introduce the supplier size there'll be more examples?

pages_builder/pages/custom-dimensions.yml
katstevens: Ooops - fixed now.

PR #409, opened 2018-02-21
Add radio button reveal

Wynndow:
### Extend `_selection-button` with conditionally revealing content

We want to use the pattern defined in GovUK pattern library for
conditionally revealing content
(http://govuk-elements.herokuapp.com/form-elements/#form-toggle-content).

This allows the options for radio selection buttons to have an extra
attribute `reveal` which defines the text box to reveal.

Laurence has approved the validation styling for the revealed component.


toolkit/scss/forms/_selection-buttons.scss
```diff
@@ -9,6 +9,9 @@
 @import "conditionals";
 @import "typography";
 @import "elements/forms/_form-multiple-choice";
+@import "elements/_helpers";
+@import "elements/_panels";
+@import "elements/_forms";
```
CrystalPea: where are these three new imports used?

Wynndow: These are scss files in govuk elements which define styles and functions used in the new html. `_panels` defines the classes used for the grey border, `_forms` defines the classes used on the form elements, and `_helpers` defines a function, `em()` which is used in those classes to defined the size of stuff.

CrystalPea:
screen shot 2018-02-22 at 14 31 57

samuelhwilliams:
Hmm, is this validation correct? I feel like we don't have the double line on followup questions 🤔

Wynndow:
I've run it passed @laurenceberry and he's happy. The grey line is there permanently to show that it's a revealed thing. Follow up questions don't have the grey, do they?

TheDoubleK:
This is how validation errors look for the G-Cloud questions with a reveal - the grey turns red, you don't get a new bar:
screen shot 2018-02-22 at 17 36 58

Wynndow:
Have update the styling. Need to run it by @laurenceberry but it now looks like the below screenshots. It also adds a hint as we do elsewhere in case of no JS (included in screenshots).

No errors:

screen shot 2018-02-23 at 11 34 20

No errors, no JavaScript (field is never hidden):

screen shot 2018-02-23 at 11 33 42

Errors, without and without JavaScript (the same):

screen shot 2018-02-23 at 11 34 00

samuelhwilliams:
Nice,  that looks a lot cleaner I think.

The content is slightly weird for the non-js hint text but it's probably livable.

Wynndow:
It's the same content we use for the other hidden inputs we have. I'll get @laurenceberry to take a look and if he's happy i'll get it in.

samuelhwilliams:
Yeah I know its the same input - but it refers to an 'earlier' question, whereas this one reveals a new question sort of _inside_ the parent. So earlier feels a bit weirder in this case. :shrug: like I said its probably fine. Just noting it.

laurenceberry:
It’s good if consistent with gcloud.

Your first version was a good solution as it didn’t rely on colour to
highlight the error, however there will be a validation banner that links
to the field directly so it is clear for all users either way.

I don’t think you need to highlight the parent. I’ll have a look at what
other services do on Monday and let you know.

laurenceberry:
Looks good y'all. 
The default non-js hint is more confusing than not having it at all in my opinion, could we drop it from this pattern?

Wynndow:
It's gone, going to merge.

PR #412, opened 2018-02-28
19px banner-content

samuelhwilliams:
## Summary
Apply core-19 to banner-content in notification banners and bump version
to 28.5.1.

The only other place banner-content is used within a notification banner is on the briefs app, when a buyer is withdrawing their brief and they get prompted to confirm the action. @laurenceberry agrees this should be updated as well, which will happen when this new version gets pulled into the relevant repo.

## Example (old)
screen shot 2018-02-28 at 17 39 38

## Example (new)
screen shot 2018-02-28 at 17 39 23

## Ticket
https://trello.com/c/JVr9gFEC/56-incomplete-application-banner


laurenceberry:
This looks good to me.

PR #413, opened 2018-02-28
flash messages template: add ability to handle flash messages of type "track-page-view"

risicle:
Tangentially related to https://trello.com/c/H7upcsyQ/303-flash-messages-suppliers-app

This is an existing mechanism found in a few places to allow analytics "page views" to be triggered based on server-side events  - it was not my idea - but at least it would be better to standardize this so that we at least gain some robustness across apps.

For more information, see https://alphagov.github.io/digitalmarketplace-manual/google-analytics.html - a page which should also be updated in accordance to this change (this change wouldn't actually break anything, but that section should be expanded to cover this new, more automatic way of handling this).

Choice of `track-page-view` as category name is up for discussion. There wasn't a previous standard naming, it was case-by-case.

Do I need to do anything more in this repo to cover this? Tests? It doesn't seem to have an entry in the `pages` to update...


risicle:
Damn I hadn't thought of that...

katstevens:
Yep, it's been moved into the FE toolkit, ready to use 😸 : https://github.com/alphagov/digitalmarketplace-frontend-toolkit/blob/master/toolkit/javascripts/analytics/_virtualPageViews.js#L16

risicle:
And actually anyway @galund, this code wouldn't have made that situation any worse - the messages would have still been consumed by another page, it just wouldn't have known to create a special `div` from it.

risicle:
@katstevens has it not only been moved to the FETK, but is the vpv js actually *run* by default by all pages?

katstevens:
Yep - we pulled it in to all the FE apps (except User FE, which basically has no templates anyway, and Admin FE, which has no analytics) for the custom dimension story.

PR #414, opened 2018-03-01
Add banner to account page

DilwoarH:
Add the user research banner to the account page using the existing GOV.uk pattern. Allow for links to 'sign up to mailing list' page

As a  user I should be presented with thte banner to sign up to the user research mailing list when I log in if: I have not already closed it and I have not already signed up for this.
This is the first point of contact for getting users to sign up to the mailing list.

The close link should set a cookie value of 'seen_this_banner_thing' to true so we don't show it again
The sign up link should link to the new sign up page.

Ticket: https://trello.com/c/OJ9lFk1M/129-add-banner-to-account-page

Documentation:
screen shot 2018-03-06 at 16 51 03


toolkit/scss/_user-research-consent-banner.scss
```diff
@@ -0,0 +1,29 @@
+.user-research-banner {
+  background-color: #005da5;
```
idavidmcdonald: Can all our colours be SASS variables please

toolkit/javascripts/user-research-consent-banner.js
```diff
@@ -0,0 +1,56 @@
+;(function (global) {
+  "use strict";
+
+  var UserResearchBanner = function( _userResearchBanner ) {
+    this.$userResearchBanner = _userResearchBanner;
+    this.$userResearchBannerCloseBtn = this.$userResearchBanner.find('.user-research-banner-close-btn');
+    this.userResearchCookie = {
+      name: 'seen_user_research_message',
```
idavidmcdonald: Nice work with consistency with the gov.uk cookies cookie

toolkit/templates/user-research-consent-banner.html
```diff
@@ -0,0 +1,11 @@
+<div class="user-research-banner">
```
idavidmcdonald: We could instead use `id` as opposed to `class` as we should only be having one of these. The GOV.UK cookie banner seems to use `id="global-cookie-message`. I don't feel too strongly on this though. Happy to go with what you feel is best.

toolkit/templates/user-research-consent-banner.html
```diff
@@ -0,0 +1,11 @@
...
+  <div class="wrapper">
+    <p>
+      <strong class="banner-heading">Help us improve the Digital Marketplace</strong>
+      <a href="#" class="user-research-banner-close-btn">Close</a>
+    </p>
+    <p>
+      <a href="#">Sign up to be a potential user research participants</a>
```
idavidmcdonald: Any idea where this link is going?

toolkit/templates/user-research-consent-banner.html
DilwoarH: I don't feel too strongly but I tend to use classes where ever possible and only IDs when actually required. Conflicting IDs can sometimes break older browsers

DilwoarH:
@idavidmcdonald - I have addressed the comments

toolkit/templates/user-research-consent-banner.html
```diff
@@ -0,0 +1,11 @@
+<div class="user-research-banner">
+  <div class="wrapper">
+    <p>
+      <strong class="banner-heading">Help us improve the Digital Marketplace</strong>
+      <a href="#" class="user-research-banner-close-btn">Close</a>
+    </p>
+    <p>
+      <a href="{{ url_for('external.user_research_consent') }}">Sign up to be a potential user research participants</a>
```
idavidmcdonald: Is the `s` on the end of the sentence correct?

DilwoarH: Good catch

toolkit/javascripts/user-research-consent-banner.js
```diff
@@ -0,0 +1,56 @@
...
+
+  UserResearchBanner.prototype.showBanner = function () {
+    this.$userResearchBanner.show();
+  };
+
+  UserResearchBanner.prototype.hideBanner = function () {
+    this.$userResearchBanner.hide();
+  };
+
+  UserResearchBanner.prototype.getCookie = function () {
```
benvand: Will this cause the banner to show if the user isn't logged in?

DilwoarH: The template should not be inserted if the user is anonymous. Not sure if there is some JS helper that allows us to see if the user is logged in or not

PR #415, opened 2018-03-07
Add hidden text attribute to summary link element for accessibility

CrystalPea:
Add hidden text attribute to summary link element for accessibility. 

I copied the existing pattern from edit_link element in the same file.


VERSION.txt
```diff
@@ -1 +1 @@
-28.6.0
+28.6.1
```
katstevens: I'd put this as a minor bump rather than a patch - we're offering new functionality.

CrystalPea: sounds reasonable 👍

PR #417, opened 2018-03-12
Bump version to 28.8.2

DilwoarH:
brings in display logic into the template to reduce duplicate code


benvand:
I think this logic might be better off in the frontend apps. We can do a conditional include there. This feels like we're moving application logic into the frontend toolkit. The frontend toolkit should really just be a collection of templates that you can include and pass simple values to.

Same with `digitalmarketplace-frontend-toolkit/toolkit/templates/proposition-header.html` which is the only other place we do it.

For proposition header it would be something like:
{% with 
  show_header=current_user.is_authenticated() if logged_in is not defined else logged_in
  role = current_user.role
  items=items
  csrf_token=csrf_token
%}
  {% include toolkit/templates/proposition-header.html %}
{% endwith %}

DilwoarH:
Closed due to not required anymore

PR #419, opened 2018-03-12
Bump version to 28.8.2

idavidmcdonald:
We had dropped IE6 support (https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/396). We had some leftover
references to `mobile-ie6` which I don't think we need as
this variable would only come into play when also supporting
IE6 and is not needed for other versions of IE. This is based
on looking at how the GOV.UK elements does not define `mobile-ie6`
for IE7/8 stylesheets and by looking at the conditionals code
that would use that variable. I believe any variations such as
`mobile-ie7` or `mobile-ie8` are not variables to use and would do
nothing.

https://github.com/alphagov/govuk_elements/tree/master/assets/sass
https://github.com/alphagov/govuk_frontend_toolkit/blob/master/stylesheets/_conditionals.scss


DilwoarH:
I'm just thinking, what is the default value of $mobile-ie6 on the GOVUK frontend? I know they've dropped support too but have we updated to their latest version of the toolkit.

idavidmcdonald:
So we are currently using v5.0.3 on the govuk frontend toolkit so this is the code we care about:

https://github.com/alphagov/govuk_frontend_toolkit/blob/v5.0.3/stylesheets/_conditionals.scss

```$mobile-ie6: true !default;```

Note, that regardless of it's value, because we won't be supporting IE6 that variable won't ever come into play as it's only used in logic statements that also require the ie-version to be 6 (which won't be the case for us anymore) - see https://github.com/alphagov/govuk_frontend_toolkit/blob/v5.0.3/stylesheets/_conditionals.scss#L34. This behaviour is the same for the version of the toolkit we are using and also the master branch.

PR #420, opened 2018-03-13
Add button and link functionality to temporary-message macro

Wynndow:
Part of this ticket: https://trello.com/c/e1Gk2eTF

We need temporary messages to include links and buttons. This updates the frontend toolkit to easily allow them.

### Screenshot of how the examples look
screen shot 2018-03-14 at 11 29 21


pages_builder/pages/temporary-message.yml
```diff
@@ -28,3 +28,25 @@ examples:
...
+    button:
+      type: save
+      label: Apply
+  -
+    heading: Your G-Cloud X application
+    main: True
+    headingId: closed-notice
+    grid: column-two-thirds
+    messages:
+      - Example of a variant with a button.
```
CrystalPea: this is a variant with the link :)

toolkit/scss/_buttons.scss
```diff
@@ -27,6 +27,10 @@ input.button-save-with-advice {
   }
 }
 
+input.button-with-temporary-message {
```
TheDoubleK: The name seems odd here - is it not more like "button-inside-temporary-message"?  And in that case, rather than applying the class to the button it would make more sense to me to add the additional margin as part of the temporary message styling, rather than as part of the button styling, the same as you've added the special link styling for links in `_temporary-message.scss`.

So have no new class name at all (i.e. no "button-with-temporary-message" class), but instead have special styling for buttons inside a temporary-message.

Wynndow:
@TheDoubleK Have moved the margin styling to be on the temporary message, let me know what you think!

TheDoubleK:
Definitely better like this.

PR #424, opened 2018-03-20
Dh 143 analytics for user research opt in

DilwoarH:


benvand:
Nice, good bit of tidying up there too 👍

PR #425, opened 2018-03-22
flash messages template: add stopgap handling of old "flag" message category

risicle:
Related to ticket https://trello.com/c/H7upcsyQ/303-flash-messages-suppliers-app

I think I've explained this enough in the included inline comment. Should be able to remove the handling of `account-created` once the user frontend no longer emits it.


toolkit/templates/flash_messages.html
```diff
@@ -3,7 +3,17 @@
...
         <div data-analytics="trackPageView" data-url="{{ message }}"></div>
-      {% elif category != 'must_login' and category != 'flag' %}  {# skip showing Flask's "Please log in" message; also skip "flag" messages intended for Google Analytics etc. #}
+      {% elif category == 'flag' %}
+        {# these were originally messages with a function similar to the above "track-page-view" ones (analytics) but #}
+        {# they did not know about their actual virtual page view url, expecting the receiving page to know the       #}
+        {# corresponding url for a particular symbolic "message".                                                     #}
+        {# TODO these can be got rid of as soon as the app generating each flag flash message is converted to using   #}
+        {# new "track-page-view" mechanism                                                                            #}
+        {% if message == "account-created" %}
+          <div data-analytics="trackPageView" data-url="/suppliers?account-created=true"></div>
```
katstevens: We have a similar `account-created` flag message for buyers (https://github.com/alphagov/digitalmarketplace-buyer-frontend/blob/master/app/templates/index.html#L17) - we should be a bit careful with hard coding the URL to `/suppliers...` here?

risicle: Well that's interesting. I didn't know the unified `flash_messages.html` could peacefully coexist with additional flash message handling - I assumed after the first `get_flashed_messages` call the messages were "consumed" and then wouldn't be seen in subsequent calls (which I *think* is the way things worked in django-land).

Anyway, :hankey: 

I'll have to fall back (*back*) to plan A and copypasta the modified `flash_messages.html` into the supplier frontend for the meantime.

risicle: Actually on that note, how far away are we from having the user-frontend converted to `track-page-view` anyway? At the rate things are going, that might get merged before the supplier frontend one and this will all be academic...

katstevens: There's a lot fewer flashes to convert in the User FE, however I haven't started it yet 😸 

Re: consuming messages - it's per request, apparently (https://github.com/pallets/flask/blob/master/flask/helpers.py#L402 "Further calls in the same request to the function will return the same messages"). So we definitely only want to have one template snippet calling `get_flashed_messages` per page.

PR #428, opened 2018-04-04
Instrument GA tracking to capture clicks made on downloads and extern…

DilwoarH:
…al links

A lot of GA tracking comes 'out of the box' but a notable exception is tracking to capture clicks on downloads and external links. Additional code is required in order to capture these.

Currently, we are tracking some downloads and external links on the DM site (those that were specifically required for previous mission work).

The proposal is to have all downloads and external links tracked automatically across the whole site, similar to how GOV.UK have done.

The advantage of doing this is that we have key data available straight away at the start of mission teams rather than having to prioritise the work, use up dev resource and then wait for enough data to populate before being able to carry out analysis.

---> includes making sure all links on the homepage are tracked, so we can measure which links are used the most and least.

Ticket: https://trello.com/c/i912cDBe/110-instrument-ga-tracking-to-capture-clicks-made-on-downloads-and-external-links


toolkit/javascripts/analytics/_trackExternalLinks.js
```diff
@@ -0,0 +1,19 @@
+(function (GOVUK) {
+  GOVUK.GDM.analytics.trackExternalLinks = {
+    'init': function () {
+      this.trackExternalLinks();
+    },
+
+    trackExternalLinks: function () {
+      $('a').filter(function () {
+        return this.hostname && this.hostname !== location.hostname;
```
katstevens: Do we want to count e.g. S3 asset links as 'external'? Document download links on a supplier service page would be an example.

katstevens: Looking at the current analytics code in the Buyer FE, we have a function to determine whether a link `isOnHostDomain`:

```
 LinkClick.prototype.isOnHostDomain = function (url) {
      var currentHost = GOVUK.GDM.analytics.location.hostname(),
          currentHostRegExp = (currentHost !== '') ? new RegExp(currentHost) : /^$/g;

      return url.match(currentHostRegExp) !== null;
 };
```
I'm not sure what's in `GOVUK.GDM.analytics.location.hostname` - should we be using this instead of `location.hostname`?

DilwoarH: that function basically just returns the same, I'll update it for consistency 
```
hostname: function() {
                return e.location.hostname
 }```

DilwoarH:
@katstevens updated

katstevens:
OK - if we want to track downloads, should that be done here too? What would happen if I clicked service document with URL https://assets.digitalmarketplace.service.gov.uk/g-cloud-9/documents/1234/terms-and-conditions.odt - would `assets.*` trigger both an external link event and a download link event?

It would be great to have some examples of what sort of link would trigger which event - perhaps in the analytics manual page: https://alphagov.github.io/digitalmarketplace-manual/google-analytics.html

DilwoarH:
@katstevens - I spoke to ash and right now we'll just focus on external links. As assets is still within the digital marketplace, we can leave it as is.

Regarding the manual - I will put in a section into the manual for tracking external links.

DilwoarH:
Hey @katstevens  - put in a small paragraph for general tracking and external links

https://github.com/alphagov/digitalmarketplace-manual/blob/external-links/google-analytics.rst#general-tracking

PR #435, opened 2018-04-25
Upgrade deps due to security warnings

Wynndow:
This is the result of me looking at the Snyk vulnerability report and having a go at following its advice.
It's advice was split into three parts. Updating versions, applying Snyk patches, and doing nothing. This PR only addresses the first part, updating versions.

It's reduced the number of reported issues from Snyk from 6 to 4. Unfortunately it only fixed the two low priority issues. The next step would be to allow Snyk to apply its patches to fix issues that can't be remedied via version bumping.

I'm unsure as to the use of this PR but thought it worth opening to see what people's feelings are. 

### Upgrade gulp to version 4.0.0  …
Snyk is reporting that we should update gulp to v4.0.0

### Remove dependency on `jasmin-core`  …
Jasmin is brought in by `gulp-jasmin-phantom`. We don't need to specify
it here.

### Fix `jasmineVerison` in `gulp-jasmine-phantom` options  …
`gulp-phantom-jasmine` takes an object of options with initialized. One
of these in the version of Jasmine to use (it supports v2.0 - v2.4). To
set this, the key used should be `jasmineVersion`, not `jasmine`. If
`gulp-phantom-jasmine` doesn't find this key, it defaults to 2.0, so
this update doesn't actually make a functional difference.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -54,9 +54,7 @@ gulp.task(
   )
 );
 
-gulp.task('copy',
-  ['copy:govuk_frontend_toolkit', 'copy:govuk-elements-sass']
-);
+gulp.task('copy', gulp.series('copy:govuk_frontend_toolkit', 'copy:govuk-elements-sass'));
 
 // Test
 
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -77,7 +77,7 @@ gulp.task('test', function () {
 
   return gulp.src(manifest.test)
     .pipe(jasmine({
-      'jasmine': '2.0',
+      'jasmineVersion': '2.0',
       'integration': true,
       'abortOnFail': true,
       'vendor': manifest.support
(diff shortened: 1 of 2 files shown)


Wynndow:
All of the frontend apps have the same vulnerabilities as the frontend toolkit, and for the same reason. I started with the frontend toolkit as I thought it might have a knock on effect if pulled into the frontend apps, but I don't think it will. They all have their own dependencies. Gulp here is just used for building/testing the pages when running locally. 

The frontend apps have their own dependencies because we use Gulp to build stuff when we developing locally, as well as when building the docker image that goes through the release pipeline, so no, we don't have to do them all at the same time.

I think phantomjs is no longer being supported so we should probably find an alternative...

brenetic:
The alternative could be gupl-jasmine-browser. Fundamentally if this fixes is up in the mean time then cool. However I thought that Gulp 4 had some changes in it that would need to be reflected in all our gulp tasks?

Wynndow:
Yeah, gulp4 is a little different. The main thing seeming to be the way tasks are defined. I've updated the gulp file in this PR to fix the breaking change. Might be more work on the frontend apps though as they have more tasks. This was quite useful though: https://www.liquidlight.co.uk/blog/article/how-do-i-update-to-gulp-4/

PR #436, opened 2018-04-27
Move buyer app templates into frontend-toolkit

DilwoarH:
Epic: https://trello.com/c/5pZdW4Cb/122-centralise-frontend-header-footer-templates

* Buyer app templates such as `_base_page.html` move into the frontend-toolkit.
* Organise the frontend-toolkit the frontend toolkit so it's easy to distinguish between its existing set of 'components' and templates like this that we are just including or extending from other DM apps;
* Move any Javascript that's easy to move but we can mostly come back to refactoring of that in another ticket.

Ticket: https://trello.com/c/lzYdOQ4C/56-move-buyer-app-templates-into-frontend-toolkit

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "digitalmarketplace-frontend-toolkit",
   "description": "A toolkit for design patterns used across Digital Marketplace projects",
-  "version": "28.13.2",
+  "version": "28.13.3",
   "repository": "git@github.com:alphagov/digitalmarketplace-frontend-toolkit.git",
   "author": "enquiries@digitalmarketplace.service.gov.uk",
   "private": true,
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "digitalmarketplace-frontend-toolkit",
   "description": "A toolkit for design patterns used across Digital Marketplace projects",
-  "version": "28.13.3",
+  "version": "28.14.0",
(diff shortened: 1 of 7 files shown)


galund:
Please make sure you pick someone from another team to do code reviews too! - thanks

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "digitalmarketplace-frontend-toolkit",
   "description": "A toolkit for design patterns used across Digital Marketplace projects",
-  "version": "28.13.2",
+  "version": "28.13.3",
```
galund: I reckon this merits at least a minor bump, it's a whole new feature for the toolkit :-)

DilwoarH: fair point 👍

DilwoarH:
@galund version number updated

DilwoarH:
@samuelhwilliams - we're moving the templates stuff over for buyer frontend - I merged prematurely (could you have a look to see if it's all fine? It's just a copy and paste job from buyer frontend)

samuelhwilliams:
Looks good to me 👍

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "digitalmarketplace-frontend-toolkit",
   "description": "A toolkit for design patterns used across Digital Marketplace projects",
-  "version": "28.13.2",
+  "version": "28.14.0",
```
samuelhwilliams: Meh, now that I think about it we probably should have done this as a breaking change to force/encourage people to move over to using the toolkit's new base layout.

PR #440, opened 2018-05-30
Allow setting input type for textbox

samuelhwilliams:
## Summary
In order to change the user-frontend over to using frontend-toolkit
textbox macros rather than manually constructing the HTML, we need to be
able to set the type of the textbox to passwor so that we can obscure
text entry in the field. This adds a kwarg to textboxes for `type` that
defaults to `text`, but can be overridden as needed.

We also need to be able to toggle whether or not the input should allow
browser to autocomplete text in the box. By default, we will disable
autocomplete for password inputs.

 ## Ticket
https://trello.com/c/UtFxXymO/455

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "digitalmarketplace-frontend-toolkit",
   "description": "A toolkit for design patterns used across Digital Marketplace projects",
-  "version": "29.0.0",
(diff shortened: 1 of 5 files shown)


samuelhwilliams:
Updated the pages builder. 👍

katstevens:
Awesome!

PR #441, opened 2018-06-01
Iterate opt in flow for user research participants

DilwoarH:
**As a buyer I want to easily opt-in to user research so that I can help improve the buyer experience for G-Cloud**

The current opt-in journey has been unsuccessful, only 6 buyers have signed up. A higher number of suppliers, however, we suspect that this is due to G10 submission going on at the same time.

Buyer and suppliers cannot opt-in until they log in. Buyers and suppliers that don't log in don't have an opportunity to opt-in for user research.

***

**Changes**

Cookie banner to appear on all pages for logged in and logged out users until they click close or opt in. If clicked close, it reappears after 90 days.

***

**Notes**

Users will have to log in before they can opt-in.
We should use the existing opt-in journey as much as possible. This is non-mission work.

Ticket: https://trello.com/c/48pZNhaG/120-iterate-opt-in-flow-for-user-research-participants

### DIFF
--- a/toolkit/templates/layouts/_base_page.html
+++ b/toolkit/templates/layouts/_base_page.html
@@ -20,6 +20,12 @@
   {% include "toolkit/proposition-header.html" %}
 {% endblock %}
 
+{% block after_header %}
+  {% if not current_user.user_research_opted_in and not 'seen_user_research_message' in request.cookies %}
+    {% include "toolkit/user-research-consent-banner.html" %}
+  {% endif %}
+{% endblock %}
+
 {% block content %}
     {% block top_header %}
     {% endblock %}
--- a/toolkit/templates/layouts/_base_page.html
+++ b/toolkit/templates/layouts/_base_page.html
@@ -21,6 +21,10 @@
 {% endblock %}
 
 {% block after_header %}
+  {# Changing cookie name will require an update to the following files: #}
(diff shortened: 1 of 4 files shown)


toolkit/templates/layouts/_base_page.html
```diff
@@ -20,6 +20,12 @@
   {% include "toolkit/proposition-header.html" %}
 {% endblock %}
 
+{% block after_header %}
+  {% if not current_user.user_research_opted_in and not 'seen_user_research_message' in request.cookies %}
```
galund: Where there's a string i.e. `seen_user_research_message` that needs to be the same in two files, and we can't define a constant somewhere, then we should probably have a comment at each end referencing the other thing? (i.e. between here and the JS file)

galund:
Dunno but this may be a breaking change, in that it mustn't go live until the banner has been _removed_ from the other places that it currently appears?

samuelhwilliams:
I can't help but slightly question the validity of the user story here...

galund:
The users need is to be able to close the box and get it out of the way, without precluding themselves from ever participating in user research. Users often _do_ want to participate, and they often _do_ want to close annoying banner messages, and there's nothing in the current UI that says that closing the box means you will never see it again.

So we're catering to some users' needs. We cannot know how many, until we see how many more people sign up as a result of this change ;-)

PR #442, opened 2018-06-04
adds custom meta data for supplier organisation size

DilwoarH:
This has been copied over from the supplier frontend to allow the app to use the generic base template


toolkit/templates/layouts/_base_page.html
```diff
@@ -1,6 +1,13 @@
 {% extends "govuk/govuk_template.html" %}
 
 {% block head %}
+  {% with custom_dimensions=[
+        {"data_id": 11, "data_value": current_user.supplier_organisation_size }
+    ] if current_user and current_user.supplier_organisation_size else []
+  %}
+    {% include "toolkit/custom-dimensions.html" %}
```
galund: Can this not move into that separate custom-dimensions.html file?

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "digitalmarketplace-frontend-toolkit",
   "description": "A toolkit for design patterns used across Digital Marketplace projects",
-  "version": "30.0.0",
+  "version": "30.0.1",
```
katstevens: I'd say this warrants a minor bump rather than a bugfix. It's a new feature we want the apps to use.

toolkit/templates/layouts/_base_page.html
```diff
@@ -1,6 +1,13 @@
...
+  {% endwith %}
+
   {% block custom_meta %}{% include "toolkit/custom-dimensions.html" %}{% endblock %}
```
katstevens: I'd rather avoid having two `custom-dimensions.html` includes in the base template if we can. 

Could we add an extra `{% with app_specific_custom_dimensions=[whatever] %}` so we can have our default ones here, and append any extra ones when we use the template in the app? 

Alternatively, if we're always going to set them here rather than in apps, why not do it in the `custom-dimensions.html` template itself, like we do with `current_user.role`?

toolkit/javascripts/report-a-problem.js
```diff
@@ -1,5 +1,5 @@
 // REDUCED VARIANT OF https://github.com/alphagov/static/blob/master/app/assets/javascripts/report-a-problem.js
-(function() {
+;(function() {
```
katstevens: Interesting! I'm not familiar with this JS syntax myself but I notice not all the `.js` files in this folder have that semi colon - when do/don't we need it?

samuelhwilliams: We should probably start+end all of our JS files with a semi colon to allow them to be safely concatenated between other JS files. At least that's the conclusion I've drawn from this fix and a quick read around.

toolkit/templates/default-custom-dimensions.html
```diff
@@ -0,0 +1,14 @@
...
+
+{# User organisation size custom dimention #}
+{% if current_user and current_user.supplier_organisation_size %}
+  <meta name="ga_customDimension" data-id="11" data-value="{{ current_user.supplier_organisation_size }}">
+{% endif %}
+
+{# Custom dimention for user research banner #}
+{% if not current_user.user_research_opted_in and not 'seen_user_research_message' in request.cookies %}
+  <meta name="ga_customDimension" data-id="12" data-value="User research banner">
+{% endif %}
```
katstevens: Let's chat about this - I think it's getting overcomplicated, given that we are only using `custom_meta` in one place (Buyer FE `templates/direct-award/view-project.html`).  We definitely don't need two templates!

toolkit/javascripts/report-a-problem.js
DilwoarH: We should really have gulp do this or actually check all files for issues

galund: This was new on me too. Definitely worth a bit of a note out. We should probably have a coding standard for JS, and maybe if we polish it enough it could be in the GDS Way.

toolkit/templates/layouts/_base_page.html
```diff
@@ -1,7 +1,8 @@
 {% extends "govuk/govuk_template.html" %}
 
 {% block head %}
-  {% block custom_meta %}{% include "toolkit/custom-dimensions.html" %}{% endblock %}
+  {% block custom_meta %}{% endblock %}
+  {% include "toolkit/custom-dimensions.html" %}
```
galund: Now that this is a thing we include from the base template, it should probably go into that `layouts` folder? Even though obviously it doesn't "lay anything out" :-o

DilwoarH:
@katstevens @galund - added the breaking change log

PR #447, opened 2018-06-14
Add `banner` option to temporary messages

Wynndow:
For [this ticket on Trello.](https://trello.com/c/LjbF4cLc)

Temporary messages had no margin at the top. This makes sense if in a
one-third column, or if there's nothing above it.

When used as a full width banner message however, if there is something
above it - such as a flash message - then there is no spacing between
them and it looks weird.

Flash messages use 15px of margin-top, so I've used the same value here.

### Before. Urgh.
localhost_buyers_direct-award_g-cloud_projects_795908519637080 ipad pro

### After. Better.
localhost_buyers_direct-award_g-cloud_projects_457356462217937 ipad pro

### DIFF
--- a/toolkit/scss/_temporary-message.scss
+++ b/toolkit/scss/_temporary-message.scss
@@ -40,6 +40,11 @@
   padding-bottom: 5px;
 }
 
+.temporary-message-banner {
+  @extend %temporary-message-section;
+  margin-top: 15px;
+}
+
 .temporary-message-heading {
   @include bold-24;
   margin: 0 0 10px 0;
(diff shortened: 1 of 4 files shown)


toolkit/scss/_temporary-message.scss
```diff
@@ -40,6 +40,11 @@
   padding-bottom: 5px;
 }
 
+.temporary-message-banner {
+  @extend %temporary-message-section;
+  margin-top: 15px;
```
samuelhwilliams: Will this give us a slightly-too-big whitespace above these banners when there aren't any flashes?

Wynndow: My thinking was that flash messages look ok with that margin, so these should too. Here's an example.
localhost_buyers_direct-award_g-cloud_projects_457356462217937 ipad pro 1

PR #450, opened 2018-06-18
Dh add form hint

DilwoarH:
Add option hint for selection buttons 
Ref: https://govuk-elements.herokuapp.com/form-elements/#form-hint-text

screen shot 2018-06-18 at 14 31 45

### DIFF
--- a/toolkit/templates/forms/_selection-button.html
+++ b/toolkit/templates/forms/_selection-button.html
@@ -20,6 +20,11 @@
         {{ option.description }}
       </p>
     {% endif %}
+    {% if option.hint %}
+      <span class="form-hint">
+        {{ option.hint }}
+      </span>
+    {% endif %}
   </label>
 </div>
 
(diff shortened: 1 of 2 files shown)


DilwoarH:
This one is the default from https://govuk-elements.herokuapp.com/form-elements/#form-hint-text

The hint class is used for "Reveal Hint" which comes with additional padding and custom styling

@samuelhwilliams

toolkit/templates/forms/_selection-button.html
```diff
@@ -20,6 +20,11 @@
         {{ option.description }}
       </p>
     {% endif %}
+    {% if option.hint %}
+      <span class="form-hint">
```
galund: FWIW this looks wrong to me - I think it should have been a `div`? (it's at the same level as a `p` which is block-level). Sorry I realise I'm too late and you'd have to refactor the tests on your other PR :(

DilwoarH: Hey @galund , 

I personally think block-level elements should not sit inside labels. It should always be a `span`.

I found this on the internet which seems to back up what I'm saying:
https://stackoverflow.com/a/4461959

The `p` tag above should probably change to a `span`

galund: Oh yeah, see what you mean. 👍

PR #455, opened 2018-07-26
Add an id to hidden fields so that the labels match correctly

lfdebrux:
For fields which are hidden behind a selection button, the label was 
associated with the div tag rather than the input tag.

This causes an issue with functional tests and also could cause an issue 
with accessability, so this commit changes the template so that the 
input field has an additional unique id that is used by it and the 
label.

I discovered this whilst working on [this ticket](https://trello.com/c/wlfBF5J0/51-buyer-fe-save-searchhtml-page-uses-blend-of-wtforms-raw-html-fe-toolkit-templates).

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "digitalmarketplace-frontend-toolkit",
   "description": "A toolkit for design patterns used across Digital Marketplace projects",
-  "version": "31.6.1",
+  "version": "31.6.2",
   "repository": "git@github.com:alphagov/digitalmarketplace-frontend-toolkit.git",
   "author": "enquiries@digitalmarketplace.service.gov.uk",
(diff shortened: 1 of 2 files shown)


risicle:
I don't imagine we have any guarantees over the page-uniqueness of these ids do we?

risicle:
Well, you can always have multiple forms on a page - and then have name collisions which are fine, but id collisions are not so fine. Nobody seems to solve this properly though so it's probably alright.

lfdebrux:
Oh, I see what you're saying. I'm not sure it's worth spending the effort to fix now though; although multiple forms on one page is a thing, I can't think of any use cases where it would be useful or workable to have a field name collision between them. If such a case does arise, we can fix the uniqueness then.

PR #457, opened 2018-08-23
Support Flask-Login 0.4.1

katstevens:
Trello: https://trello.com/c/euSnpBnN/65-update-flask-login

`current_user.is_authenticated()` is now an attribute, not a method (see Flask-Login Changelog: https://github.com/maxcountryman/flask-login/blob/master/CHANGES). This means the FE toolkit's `proposition-header.html` template breaks when it tries to call it.

Happily Jinja can check whether the object is a callable and evaluate it accordingly: http://jinja.pocoo.org/docs/2.10/templates/#callable

Also did some whitespace tidying so the multiple `if..else`s are a bit clearer.

### DIFF
--- a/toolkit/templates/proposition-header.html
+++ b/toolkit/templates/proposition-header.html
@@ -9,8 +9,13 @@
         <li>
           <a href="{{ url_for('external.help') }}">Help</a>
         </li>
-        {% set current_user_is_authenticated = current_user.is_authenticated() if current_user.is_authenticated is callable else current_user.is_authenticated %}
-        {% if (current_user_is_authenticated if logged_in is not defined else logged_in) %}
+        {% set user_logged_in = logged_in|default(false, true) %}
+
+        {% if current_user %}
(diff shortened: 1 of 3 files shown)


toolkit/templates/proposition-header.html
```diff
@@ -9,26 +9,32 @@
         <li>
           <a href="{{ url_for('external.help') }}">Help</a>
         </li>
-        {% if (current_user.is_authenticated() if logged_in is not defined else logged_in) %} {% if (role or current_user.role) == 'buyer' %}
```
lfdebrux: If I understand this line correctly, the is_authenticated flag is only used if logged_in is not already present, so instead of changing the pages builder template, could we instead just make the new assignment line also use logged_in if it's defined?

katstevens: I thought of an even better way, using `default` http://jinja.pocoo.org/docs/2.10/templates/#default

lfdebrux: Oooh

PR #459, opened 2018-10-01
Shared footer SCSS

katstevens:
Trello: https://trello.com/c/Wnrvezi2/174-admin-404-error-template-footer-looks-weird

All our frontend apps (except the Admin FE) use the same footer layout and content, from the shared `_base_page.html` and `_footer_categories.html` toolkit templates. 

The Admin FE doesn't show `_footer_categories.html` for its normal pages, but its (shared) error pages do show them. However the Admin FE doesn't have any footer category CSS, so the error page footers look weird (hence this ticket).

Rather than copy over the duplicated `_footer.scss` to the Admin, I thought it would be nice to centralise it in the toolkit.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "digitalmarketplace-frontend-toolkit",
   "description": "A toolkit for design patterns used across Digital Marketplace projects",
-  "version": "31.8.0",
+  "version": "31.9.0",
   "repository": "git@github.com:alphagov/digitalmarketplace-frontend-toolkit.git",
   "author": "enquiries@digitalmarketplace.service.gov.uk",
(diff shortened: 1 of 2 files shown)


Wynndow:
If the current frontend apps are using shared templates from the toolkit, where are they getting their CSS from?

katstevens:
They are all using a duplicated `_footer.css` in the individual apps! I checked them and they're all the same. Madness.

Wynndow:
Ah cool, so they'll all use this new shared stuff. 👍

katstevens:
Yep - and because it's identical, if there's some cross over (i.e. if we forget to remove it) then it doesn't matter.

PR #462, opened 2019-01-07
Update requests

lfdebrux:
Once this is merged the [GitHub alert](https://github.com/alphagov/digitalmarketplace-frontend-toolkit/network/alert/pages_builder/requirements.txt/requests/open) can be dismissed

### DIFF
--- a/pages_builder/requirements.txt
+++ b/pages_builder/requirements.txt
@@ -1,7 +1,7 @@
 PyYAML==3.11
 libsass==0.11.1
 pystache==0.5.4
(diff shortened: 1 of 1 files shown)


jonodrew:
Do we need to bump the version in `package.json` ?

lfdebrux:
> Do we need to bump the version in `package.json` ?

No, requests is only used by the `pages_builder` which is a utility that uses the library but not part of the library proper.

jonodrew:
Then looks good to me!

PR #463, opened 2019-02-08
Fix contact details examples

lfdebrux:


pages_builder/pages/contact-details.yml
```diff
@@ -8,16 +8,14 @@ examples:
     contact_name: Jo Johnston
     telephone: 012345 678910
     email_address: jo@example.com
-    street_address:
-      - line_1: Aviation House
-      - line_2: 125 Kingsway
+    street_address_line_1: Aviation House
+    street_address_line_2: 125 Kingsway
```
katstevens: I think this needs to be indented under `street_address` for the template to pick it up.

lfdebrux: Hmm, I think it might actually be the template that is broken; the [logic](https://github.com/alphagov/digitalmarketplace-frontend-toolkit/blob/master/toolkit/templates/contact-details.html#L38-L50) goes if `street_address`, then put values from `street_address_line_1` etc., so to get the expected output the YAML has to read

```
street_address: true
street_address_line_1: Aviation House
street_address_line_2: 125 Kingsway
```

which I don't think was the original intention...

katstevens: We use both in the supplier FE here: https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/master/app/templates/suppliers/details.html#L103

lfdebrux: Beautiful. As that seems to be the only place where street address is used with the contact details template, how would you feel about me changing the template to not require `street_address = True`?

katstevens: Why not. Remember to add a version bump though!

lfdebrux: 👍

PR #466, opened 2019-04-11
Upgrade FE toolkit dependencies

katstevens:
Trello: https://trello.com/c/nLGfr8un/445-frontend-toolkit-is-pinned-to-node-6

Way back in the mists of time, I tried upgrading gulp to v4 here as well (see https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/461), eventually closing the PR due to a vulnerability. This vulnerability is now fixed! So let's try again with:

- bump `node` to v8
- bump `gulp` to v4 (and update the `gulpfile` to work with v4)
- bump `jasmine-core` and `del` to most recent versions
- include my nice `make test` command and put it in travis.yml

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,6 +16,9 @@ npm_copy:
 npm_clean:
 	npm run clean
 
+test: npm_install
+	npm test
+
 generate_pages: requirements npm_install npm_clean npm_copy
 	${VIRTUALENV_ROOT}/bin/python pages_builder/generate_pages.py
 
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -16,6 +16,9 @@ npm_copy:
 npm_clean:
 	npm run clean
 
+test: npm_install
+	npm test
```
lfdebrux: 👍

PR #469, opened 2019-05-30
479-spike-4h-back-button-does-not-work-as-expected-on-search-page-2

benvand:
Mitigation for browser versions that call popstate on pageload as opposed to just browser back/ forward events


benvand:
[Back button does not work as expected on search page](https://trello.com/c/kcElJmwF/479-back-button-does-not-work-as-expected-on-search-page)

PR #471, opened 2019-06-13
534-add-resize-polyfill-to-frontend-toolkit

benvand:
https://trello.com/c/BSllYDtx/534-add-resize-polyfill-to-frontend-toolkit

A rework of https://github.com/CezaryDanielNowak/css-resize-polyfill/blob/master/lib/polyfill-resize.js

Adds resize functionality to IE11

*Before:*

Screen Shot 2019-06-13 at 13 06 11

*After:*

Screen Shot 2019-06-13 at 13 05 14

### DIFF
--- a/toolkit/templates/forms/textbox.html
+++ b/toolkit/templates/forms/textbox.html
@@ -64,24 +64,28 @@
     {% if large %}
       {% if max_length_in_words %}
         <div class="word-count">
-          <textarea
-            class="text-box text-box-large{% if error %}-with-error{% endif %}"
-            name="{{ name }}"
-            id="input-{{ name }}"
-            data-max-length-in-words="{{ max_length_in_words }}"
-            {% if question_advice or hint %}aria-describedby="{{ [question_advice_id, answer_advice_id]|join(' ')|trim }}"{% endif %}
-          >
+          <div>
+            <textarea
+              class="text-box text-box-large{% if error %}-with-error{% endif %}"
+              name="{{ name }}"
+              id="input-{{ name }}"
+              data-max-length-in-words="{{ max_length_in_words }}"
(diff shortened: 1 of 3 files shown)


toolkit/templates/forms/textbox.html
```diff
@@ -64,24 +64,28 @@
...
       {% if max_length_in_words %}
         <div class="word-count">
-          <textarea
-            class="text-box text-box-large{% if error %}-with-error{% endif %}"
-            name="{{ name }}"
-            id="input-{{ name }}"
-            data-max-length-in-words="{{ max_length_in_words }}"
-            {% if question_advice or hint %}aria-describedby="{{ [question_advice_id, answer_advice_id]|join(' ')|trim }}"{% endif %}
-          >
+          <div>
```
lfdebrux: Why is the extra div required?

benvand: https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/471/files#diff-6a148e617c7e875e3f54f660a13ae54dR13

https://github.com/CezaryDanielNowak/css-resize-polyfill/blob/master/demo.html#L37

benvand: It operates on (applies classes to) an empty parent div

lfdebrux: It doesn't work on the existing parent div?

benvand: Yeah, but it felt a bit dangerous doubling up, don't want to override anything

lfdebrux: I'm just a bit worried about how multiple divs might interfere with a screenreader; it would be good to test this branch in the empathy lab

benvand:
@lfdebrux Checked with and without extra div, with and without polyfill. Screenreader behaviour the same.

(Used ngrok in the end)

PR #472, opened 2019-06-26
Strip tags from validation message bookmark links

katstevens:
https://trello.com/c/hJQFvCUe/569-dos4-service-question-error-banner-breaks-with-question-containing-link

Some of our form validation messages replay the question text as a bookmark link, to help the user jump to the affected area quickly. If the question text itself contains a link, the bookmark breaks.

Stripping the tags from the question text solves the issue. Other masthead/flash messages (that may contain links) should not be affected - this will only be used for form validation errors containing content loader questions.

Fixed error message:
validation-masthead-links

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "digitalmarketplace-frontend-toolkit",
   "description": "A toolkit for design patterns used across Digital Marketplace projects",
-  "version": "33.0.1",
+  "version": "33.0.2",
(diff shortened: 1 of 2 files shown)


lfdebrux:
Could we also add a filter so that the question is cut at the end of the first sentence? These validation messages are a bit dense...

katstevens:
@lfdebrux I'd rather change the question itself and move the second sentence to hint text.

PR #477, opened 2019-09-06
Replace yarn with npm

lfdebrux:
We do not need to depend on Yarn as npm is Nodes default package manager and we use npm to install yarn. npm does everything yarn does now.

This commit mainly updates the manual but it also replaces .yarnrc with .npmrc.

Trello ticket: https://trello.com/c/RrpjQDJi/278-replace-yarn-with-npm

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,11 +1,15 @@
-.DS_Store
+# NodeJS stuff
+node_modules
+npm-debug.log
+
+# Python stuff
 *.pyc
 __pycache__
 venv
-npm-debug.log
+
+# App stuff
 govuk_template
 /pages
 gh-pages
-node_modules
 pages_builder/govuk_frontend_toolkit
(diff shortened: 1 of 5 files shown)


.gitignore
```diff
@@ -1,11 +1,17 @@
-.DS_Store
+# NodeJS stuff
+node_modules
+npm-debug.log
+package-lock.json
+yarn.lock
```
aliuk2012: We could remove this

.gitignore
```diff
@@ -1,11 +1,17 @@
-.DS_Store
+# NodeJS stuff
+node_modules
+npm-debug.log
+package-lock.json
```
aliuk2012: We want to track package-lock.json.

lfdebrux:
> Missing a package-lock.json file. If you run `npm install` it should generate one for you to commit.

Good spot. Added now.

PR #478, opened 2019-09-06
base template link updates

risicle:
https://trello.com/c/lxumtOjc

I am under the command of Stefan.


risicle:
Yaaaay oh now I've got to think about omnibumping and simultaneous release...

PR #483, opened 2019-10-24
Removes standard 400 & 500 templates

aliuk2012:
We have moved the 400 & 500 error templates to a new repo.

https://github.com/alphagov/digitalmarketplace-govuk-frontend/pull/14


katstevens:
Can we make sure this gets captured in the Changelog/version number before we merge to master 👍

PR #484, opened 2019-10-24
Remove GA verification and custom dimensions

aliuk2012:
We moved these to a new repo
https://github.com/alphagov/digitalmarketplace-govuk-frontend/pull/15


katstevens:
This will be a breaking change, so we'll need to add a major version bump & update the Changelog so we know not to pull it in to apps that haven't been upgraded yet.

PR #487, opened 2019-11-19
Removes 🤢 Hogan.js

aliuk2012:
Hogan has not been maintained or released for over 5 years. Looking at what we are using Hogan for, it seems a bit overkill for templating and we could achieve the same results using find/replace.

- Removes Hogan dependency from ListEntry.js
- Removes any reference to Hogan.js
- Fixes a minor usability issue identified whereby users could not remove the first entry even if there was more than one entry in the list. Users will not be able to remove the last entry if there is only one entry in the list.

## Before:
image

## After:
image


aliuk2012:
@tombye Would you mind reviewing this from Frontend perspective?

README.md
```diff
@@ -92,7 +92,7 @@ When writing SASS for this toolkit, [follow the styleguide](toolkit/scss/README.
 Javascript modules should be initialised by calling them from
 `./pages_builder/assets/javascripts/onready.js`.
 
-jQuery and Hogan are made available for you.
+jQuery is available for you to use but only as last resort.
```
tombye: Might be worth linking to http://youmightnotneedjquery.com/ here.

katstevens:
Agree that this probably isn't a breaking change, as the frontend apps have `hogan.js` as a top level dependency in `package.json` anyway. 

Speaking of which, is there a separate ticket to remove `hogan.js` from the `package.json` and `application.js` on each of the FE apps? Or will it be done as part of this one?

aliuk2012:
@katstevens I haven't created another ticket but I think that we should remove hogan.js from package.json and application as part of updating each app to this new version.

aliuk2012:
I will create a ticket for that piece of work

PR #494, opened 2021-01-22
Change single field pricing question to also look for maximum price

alex9smith:
Before DOS5 all pricing questions with a single field were on G-Cloud and the field was `price`. In DOS5 we switched to asking for only a specialist's maximum day rate, rather than both a maximum and minimum rate. This means we now have single field pricing questions with fields `price` and `maximum_price`.

The field name is used to populate the form input name, for questions with `maximum_price` this attribute was empty. This caused a bug in the admin-frontend where an admin user could not update a DOS5 specialists maximum day rate.

Update the name attribute on single field questions to look for `maximum_price` if `price` is not present. I initially felt a bit bad about hard-coding the second option, but logically it does make sense that a single field pricing question is either asking for a service's price (G-Cloud) or maximum price (DOS specialists).
I have checked the other questions in DOS and G-Cloud and verified that these are the only two fields that appear on their own.

https://trello.com/c/XGAEqKRu/2165-ccs-category-admins-cant-add-new-specialist-types-to-digital-specialists-services-on-dos5

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "digitalmarketplace-frontend-toolkit",
-  "version": "36.1.0",
+  "version": "36.1.1",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "digitalmarketplace-frontend-toolkit",
   "description": "A toolkit for design patterns used across Digital Marketplace projects",
-  "version": "36.1.0",
+  "version": "36.1.1",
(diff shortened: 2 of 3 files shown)


bjgill (approved):
I agree with your reasoning. I assume it's OK for the value to still be `price`?

Otherwise looks good.

alex9smith:
I've had a look at updating the value but I think I'd need to make changes to the macro to get a previously submitted answer, and that'll open a whole can of worms for backwards compatibility. I'd rather leave it as is and deal with not having a prefilled field if it becomes a problem. I think quite a lot of this changes when we move to design system 3 anyway, so it might not be a problem for long.

PR #496, opened 2021-01-26
Add merge actions

gidsg:
https://trello.com/c/YXlsZjTP/745-publish-frontend-toolkit-using-github-actions-05
Adds Github Actions to:
* Publish Github Pages
* Publish Github Release/Tag
* Publish to NPM

This is intended to superceed the actions previously performed by Jenkins i.e. https://github.com/alphagov/digitalmarketplace-jenkins/blob/master/job_definitions/toolkit.yml
and additionally start publishing to NPM to allow for dependabot versioning.

This shares most logic with the existing similar actions used by [manual](https://github.com/alphagov/digitalmarketplace-manual/blob/master/.github/workflows/manual.yml) and [test-utils ](https://github.com/alphagov/digitalmarketplace-test-utils/tree/main/.github/workflows)

### Note on testing

I have tested the publish action on a fork but I haven't tested creating to releases or publishing yet. I added a commit to create a beta release and this would get tested on merge, at which point that commit can be reverted.


.github/workflows/publish_package.yml
```diff
@@ -0,0 +1,18 @@
+name: Node.js Package
+on:
+  release:
+    types: [published]
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    # Setup .npmrc file to publish to npm
```
lfdebrux, line 10: I don't understand this comment

.github/workflows/publish_pages.yml
```diff
@@ -0,0 +1,46 @@
...
+    steps:
+    - uses: actions/checkout@v2
+      
+    - uses: actions/setup-node@v2
+      with:
+        node-version: 10.x
+    
+    - uses: actions/setup-python@v2
+      with:
+        python-version: '3.6.x'
```
lfdebrux, line 20: Can we add `.node-version` and `.python-version` files (or equivalents) instead of hardcoding the versions in the actions?

.github/workflows/publish_package.yml
gidsg: Ah, I took this from the boilerplate, see: https://docs.github.com/en/actions/guides/publishing-nodejs-packages#publishing-packages-to-the-npm-registry

lfdebrux: I see. So the step does create an `.npmrc` file... I wonder how it does that. We already have an `.npmrc` file in this repo, hopefully that won't interfere with this step.

.github/workflows/publish_pages.yml
gidsg: @lfdebrux I have addressed your comments now, please may you re-review? Note: for python I thought it simpler to read the file in rather than using a 3rd party action but let me know if you feel strongly against this.

lfdebrux (approved):

.github/workflows/publish_pages.yml
lfdebrux: Works for me

PR #515, opened 2022-03-10
changes done for updating minimum price

pradeepkumar-bc:
changes done for pricing update issue for research studios lot


pradeepkumar-bc:
File pricing.html was allowing only maximum price, when there is a single field pricing input. But in DOS lot 2 research studio price is a single and minimum price field. Now, the code is updated to accept both the minimum and maximum price.

tberey (approved):

